PR #8, opened 2018-07-29
Add documentation on caddy

Wqrld:
I thought this would be useful, suggestions are welcome.

### DIFF
--- a/panel/webserver_configuration.md
+++ b/panel/webserver_configuration.md
@@ -50,3 +50,49 @@ sudo ln -s /etc/apache2/sites-available/pterodactyl.conf /etc/apache2/sites-enab
 sudo a2enmod rewrite
 systemctl restart apache2
 ```
+
+## Caddy
+Caddy is a webserver with automatic https, this may come handy when you don't want to remake your ssl certificates every 90 days.
+::: danger
+Caddy is incompatible with other webservers, if you plan on running other websites on your panel you will have to add them to caddy too.
+:::
+/etc/caddy/Caddyfile:
+``` json
+panel.yourdomain.com {
+    # Set the directory root
+    root /var/www/pterodactyl/public
+    # Let caddy know where fastcgi is for php, edit "7.2" to your php-fpm version
+    fastcgi / /run/php/php7.2-fpm.sock php
+    # rewrite all requests to index.php, this is required for how pterodactyl works
+    rewrite {
+        to {path} {path}/ /index.php
+}
+```
+to turn it on:
+``` bash
+caddy -conf Caddyfile
+```
+To run caddy in the background use systemD
+``` bash
+
+cat <<EOF > /etc/systemd/system/caddy@.service
+[Unit]
+Description=Caddy HTTP/2 web server %I
+Documentation=https://caddyserver.com/docs
+After=network.target
+
+[Service]
+User=%i
+Environment=STNORESTART=yes
+ExecStart=/usr/sbin/caddy -agree=true -conf=/etc/caddy/Caddyfile
+Restart=on-failure
+
+[Install]
+WantedBy=multi-user.target
+EOF
+```
+reload the systemd daemon: `systemctl daemon-reload`
+start the service: `systemctl start caddy@root`
+enable it (to start it automatically on boot): `systemctl enable caddy@root`
+to restart it: `systemctl restart caddy@root`
+For more documentation on how to use the caddy webserver, go here https://caddyserver.com/docs.
--- a/panel/webserver_configuration.md
+++ b/panel/webserver_configuration.md
@@ -83,7 +83,7 @@ After=network.target
 [Service]
 User=%i
 Environment=STNORESTART=yes
-ExecStart=/usr/sbin/caddy -agree=true -conf=/etc/caddy/Caddyfile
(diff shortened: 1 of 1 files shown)


panel/webserver_configuration.md
```diff
@@ -50,3 +50,48 @@ sudo ln -s /etc/apache2/sites-available/pterodactyl.conf /etc/apache2/sites-enab
...
+/etc/caddy/Caddyfile:
+``` json
+panel.yourdomain.com {
+    # Set the directory root
+    root /var/www/pterodactyl/public
+    # Let caddy know where fastcgi is for php, edit "7.2" to your php-fpm version
+    fastcgi / /run/php/php7.2-fpm.sock php
+    # rewrite all requests to index.php, this is required for how pterodactyl works
+    rewrite {
+        to {path} {path}/ /index.php
```
DaneEveritt: This seems to be missing a closing brace?

panel/webserver_configuration.md
```diff
@@ -50,3 +50,48 @@ sudo ln -s /etc/apache2/sites-available/pterodactyl.conf /etc/apache2/sites-enab
...
+}
+```
+to turn it on:
+``` bash
+caddy -conf /etc/caddy/Caddyfile
+```
+To run caddy in the background use SystemD
+``` bash
+cat <<EOF > /etc/systemd/system/caddy@.service
```
DaneEveritt: Is this actually the correct syntax for writing to the file? It seems incorrect.

panel/webserver_configuration.md
```diff
@@ -50,3 +50,48 @@ sudo ln -s /etc/apache2/sites-available/pterodactyl.conf /etc/apache2/sites-enab
...
+User=%i
+Environment=STNORESTART=yes
+ExecStart=/usr/local/bin/caddy -agree=true -conf=/etc/caddy/Caddyfile
+Restart=on-failure
+
+[Install]
+WantedBy=multi-user.target
+EOF
+```
+reload the systemd daemon: `systemctl daemon-reload`
```
DaneEveritt: Please clean up these instructions, they do not match the rest of the documentation, and a majority of them are unnecessary. We do not teach people how to use dependencies in this documentation, so things like how to restart the webserver or start it shouldn't be included unless they are relevant to the process itself.

panel/webserver_configuration.md
Wqrld: this was way you could just fill it into your bash shell, i've changed it to what should be in the file.

panel/webserver_configuration.md
Wqrld: yes, missed that while copying from my own setup

panel/webserver_configuration.md
```diff
@@ -50,3 +50,47 @@ sudo ln -s /etc/apache2/sites-available/pterodactyl.conf /etc/apache2/sites-enab
...
 systemctl restart apache2
 ```
+
+## Caddy
+Caddy is a webserver with automatic https, this may come handy when you don't want to remake your ssl certificates every 90 days.
+::: danger
+Caddy is incompatible with other webservers, if you plan on running other websites on your panel you will have to add them to caddy too.
+:::
+/etc/caddy/Caddyfile:
+``` json
```
DaneEveritt: This isn't JSON, just do the backticks without anything since its basically just curly text.

panel/webserver_configuration.md
```diff
@@ -50,3 +50,47 @@ sudo ln -s /etc/apache2/sites-available/pterodactyl.conf /etc/apache2/sites-enab
...
+    rewrite {
+        to {path} {path}/ /index.php
+            }
+}
+```
+to turn it on:
+``` bash
+caddy -conf /etc/caddy/Caddyfile
+```
+To run caddy in the background use SystemD
```
DaneEveritt: Can you word this like the nginx section in terms of where to place the file at? (also, not bash below, just text)

DaneEveritt:
@parkervcp could you take a peek at this, I believe you have experience with caddy. Looks fine to me, but I've never used it.

panel/webserver_configuration.md
```diff
@@ -50,3 +50,54 @@ sudo ln -s /etc/apache2/sites-available/pterodactyl.conf /etc/apache2/sites-enab
...
+WantedBy=multi-user.target
+```
+
+```
+# reload the systemd daemon:
+systemctl daemon-reload
+# start the service:
+systemctl start caddy@root
+# enable it (to start it automatically on boot):
+systemctl enable caddy@root
```
schrej: Why would you be starting caddy as root 🤔

parkervcp:
the config part should be a snippet instead of being in the doc. Match the rest at least. I will look at the config later when I can though.

panel/webserver_configuration.md
Wqrld: This looked the best to me as it seems to need a user and there is no good default username on linux as far as i am aware.
(didn't want to have an automatic $current-user thing)

lancepioch: IIRC Apache and Nginx both create their own users to run the web server as.

Maybe just create the user for them `useradd -M -s /bin/false caddy`

PR #9, opened 2018-07-29

Luukth:
When NGINX is active letsencrypt won't work, unless we used the --nginx tag when creating the certificate. With this NGINX will be disabled when creating the certificate.

### DIFF
--- a/tutorials/creating_ssl_certificates.md
+++ b/tutorials/creating_ssl_certificates.md
@@ -30,5 +30,7 @@ You can add the command below to that crontab. For advanced users, we suggest in
 which provides more options, and is much more powerful than certbot.
 
 ``` text
+service nginx stop
 letsencrypt renew
+service nginx start
 ```


Wqrld:
I also always need to do this, good to have it in the docs

tutorials/creating_ssl_certificates.md
```diff
@@ -30,5 +30,7 @@ You can add the command below to that crontab. For advanced users, we suggest in
 which provides more options, and is much more powerful than certbot.
 
 ``` text
+service nginx stop
```
schrej: Please remember that not everybody uses nginx. Rather add a warning that tells people that they might need to turn off their webserver to get certificates. The syntax for warnings is:
```
::: warning
Some warning text.
:::
```

Luukth: Yes that would also work, it's still a good idea in my opinion to give an example on how it could look with apache and nginx.

DaneEveritt:
Thanks for your PR, however none of the changes provided are necessary and simply go in depth teaching users how to use their server, which is not the goal of this documentation. Everything added here should be apparent based on the instructions given (and the errors thrown if not followed).

Wqrld:
I don't feel like that is true as i have had multiple people ask be about why it didn't work but it's up to you.

Luukth:
@DaneEveritt well.. yes I agree with that. Users that want to use this function should be able to debug it themself.

DaneEveritt:
If you don't know how to start and stop your webserver, and cannot take < 5 minutes to google the potential LetsEncrypt error and see all of the other answers online from people with the same problem, you shouldn't be using this software.

PR #23, opened 2018-08-26
Add further explanation for Gmail SMTP

Silgica:
Added info about how to use Gmail's SMTP host in Ptero correctly, thought it might go well in the tutorials section.


lancepioch:
I'm not sure if we want to add a whole page just for one email provider. If anything changes, we have to keep updating our own pages instead of users just looking up the correct and latest instructions on Gmail's website itself. Also when do we draw the line, there are tens of popular SMTP mail providers, we can't have a page for each one.

Silgica:
That's understandable, I thought it might be useful to have there anyway due to the reference to gmail's SMTP host in the setup, but I understand what you mean.

DaneEveritt:
Thanks, but yeah this is getting way too specific and isn't something I think the docs really need to cover.

PR #24, opened 2018-08-31
Shared folder docker guide

xion87:
I added this guide for help user how to share easy the folders between server, (until will be implemented in future) hope is usefull :)


lancepioch:
We appreciate the contribution, but I'm sure you can imagine the headaches involved with encouraging users to edit the source files directly.

A better solution might be allowing the users to edit a `custom-mounts.json` file that will be pulled into docker's bind mounts.

DaneEveritt:
I like that idea @lancepioch. Should resolve the issue that is open with people wanting custom mount options as well in the Daemon.

PR #26, opened 2018-09-11
Update introduction.md

DrZoddiak:
Remove Terraria from Core supported games list.

### DIFF
--- a/project/introduction.md
+++ b/project/introduction.md
@@ -12,7 +12,6 @@ Some of our core supported games include:
 
 * Minecraft — including Spigot, Sponge, Bungeecord, Waterfall, and more
 * Rust
-* Terraria
 * Teamspeak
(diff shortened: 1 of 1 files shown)


lancepioch:
Terraria is supported though: https://github.com/pterodactyl/panel/blob/develop/database/seeds/eggs/terraria/egg-terraria-server--t-shock.json

DrZoddiak:
Super apologize. I guess it hadn't been seeded on my panel at the time.

I thought it was odd that Terraria had appeared twice on the list being core and egg from community. Perhaps the docs could suggest that the available community egg is tshock?

schrej:
The docs shouldn't include information about community eggs (or themes/plugins etc.) in my opinion. The Readme of the community eggs repo could indicate it though.

PR #27, opened 2018-09-15
adding community section

parkervcp:
Sorry Dane for the config.js mess but it was necessary for now.

Transferred guides from readme to the new docs site.

PR #28, opened 2018-09-15
Redis on CentOS service name

itsgatto:
The redis service on CentOS is called just redis.service, so we need to point that out as i have seen some people having issues with the queue worker trying to start before redis and failing to do so.


parkervcp:
@itsgatto Check the #27 PR.

lancepioch:
@parkervcp I think if you add @itsgatto's changes to your PR, then we can close this one.

lancepioch:
I see your changes are now in #27, thanks!

PR #30, opened 2018-09-23
Add a fix for 18.04

BrightSkyz:


DaneEveritt:
For what reason is this being added?

BrightSkyz:
The libzip package isn't included in the default repositories on a fresh install of Ubuntu 18.04. The package is provided in the universe repository.

PR #33, opened 2018-09-27
Small changes

Ethereal739:
Adding VMware as a supported virtualization method and removing Nodejs 10 as it is not working with the daemon yet. Node-gyp should also be installed with Nodejs.


DaneEveritt:
As of the last release Node 10 is supported.

PR #37, opened 2018-10-01
Update getting_started.md

Ethereal739:
- Adding the -y tag to some commands to make the docs more consistent
- Forcing IPv4 on apt as it does not work well with IPv6 at all.


Ethereal739:
Apt on ipv6 is extremely slow and doesn't play nice at all. I had a few dedicated servers that do this and I had to force them to use IPv4.

DaneEveritt:
Forcing IPv4 is outside the scope of what this documentation is trying to do. Also, in the future please limit PRs to one area of change, and not two unrelated changes.

PR #39, opened 2018-10-04
Adding npm audit fix

Ethereal739:
Adding npm audit fix to fix security issues.

### DIFF
--- a/daemon/installing.md
+++ b/daemon/installing.md
@@ -126,12 +126,16 @@ The next step is to download the software and unpack the archive.
 curl -L https://github.com/pterodactyl/daemon/releases/download/v0.6.7/daemon.tar.gz | tar --strip-components=1 -xzv
 ```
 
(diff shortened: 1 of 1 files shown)


daemon/installing.md
```diff
@@ -126,12 +126,16 @@ The next step is to download the software and unpack the archive.
...
 ```
 
-Finally, we need to install the dependencies that allow the Daemon to run properly. This command will most likely
+Next, we need to install the dependencies that allow the Daemon to run properly. This command will most likely
 take a few minutes to run, please do not interrupt it.
 ``` bash
 npm install --only=production
 ```
 
+Finally, we will improve the security of the installation.
```
schrej: Please actually describe what the command does, I had to google it for instance.

DaneEveritt:
This is something handled at the project packaging level, and is by no means something regular users should be running unless they actually understand what it does, and can test that things aren't broken by updating packages.

PR #50, opened 2018-11-15
CentOS7 requires node8, fails on 10

Sir3lit:
Node 10 fails to install mmmagic on CentOS7, requires node 8.x


parkervcp:
I just installed the daemon on  node 10 with no issues. I am going to close this out.

PR #51, opened 2018-11-17
Errors

BLEEPBLOOPforLife:
Some people experience an error where they cannot clear their template cache because some dependencies were out-of-date.


Sir3lit:
Please use the issue template

DaneEveritt:
This should never happen under normal circumstances. Dependencies are locked to specific versions defined by (mostly) myself and my dev environment. Do you have a reproducible example of this? If so please report it on the panel repo and we can come back to making docs changes if so.

PR #52, opened 2018-11-26
Added node-gyp

Sir3lit:
Added node-gyp to dependencies

### DIFF
--- a/daemon/installing.md
+++ b/daemon/installing.md
@@ -47,6 +47,7 @@ Pterodactyl's Daemon requires the following dependencies be installed on your sy
 
 * Docker
 * Nodejs (`8.0` or `10.0`)
+* `node-gpy`
 * `tar`
 * `unzip`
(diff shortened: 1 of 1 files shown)


lancepioch:
@Sir3lit Can you post your full output showing your input and the respective output that requires this change? Thanks!

Sir3lit:
I don't understand why you need that, get a fresh install of Ubuntu literally follow the guide as you would and when you get to npm install if fails because node-gyp isn't installed

lancepioch:
@Sir3lit I don't get that. Here's my output (this is a fresh Ubuntu instance):

image

parkervcp:
I have had no issues with installing the daemon on Ubuntu either. We just want to see what exactly is failing for you.

DaneEveritt:
I'm fine adding this, but it is typo'd

Sir3lit:
Fixed

PR #55, opened 2018-12-16
add more visible warning regarding app_key backup

schrej:


panel/getting_started.md
```diff
@@ -111,6 +110,11 @@ composer install --no-dev --optimize-autoloader
 php artisan key:generate --force
 ```
 
+::: danger
+Back up your encryption key (APP_KEY in the `.env` file). It is used as an encryption key for all sensible data (e.g. passwords).
```
DaneEveritt: This is kinda correct, but misleading. It is for encrypting data that we need to store securely but still need to be able to read later on, such as secret keys.

Need to make sure this isn't worded in a way that implies we use encryption (vs. hashing) for user passwords and one-way data.

PR #59, opened 2019-01-03
Update CentOS 7 tutorial

JhonatanPereira:
The instructions about where the contents described should be placed were not clear, leading to error 403 at the end of the installation.

### DIFF
--- a/community/install/panel/centos7.md
+++ b/community/install/panel/centos7.md
@@ -115,9 +115,8 @@ We have a tutorial in the tutorial section
 
 ### PHP
 
-The default php-fpm configuration is good to use but you need to edit the nginx config to use the ip address.
-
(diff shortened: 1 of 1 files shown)


parkervcp:
The conf file can literally be named whatever they want to name it. I avoid putting `www.` and `.com` in my config file names.

JhonatanPereira:
I understand, but maybe the people who are more step-by-step oriented can be confused, since the old text does not say explicity to save the contents in any file.
I put the file name following the pattern of the default file already present in the folder, but maybe we can leave something like "You can name the file however you want, but if you don't want to bother, just use 'www-pterodactyl.conf'". What do you think?

parkervcp:
We can skip the "if you don't want to bother". Something like 'The file can be named anything but a good standard is `www-pterodactyl.conf`.'

PR #60, opened 2019-01-15
Install curl before installing mysql via curl

PiggyPiglet:
Should be fairly self explanatory by looking at the file change.


schrej:
As we have it in one of the install commands already I'm fine with moving it, even though people should be able to install it themselves when it's missing.

PR #61, opened 2019-01-16
Added documentation for acme.sh

ckykenken:
Making use of APIs of DNS providers, auto-renewal can be done even port 80 is not available on the node to the internet.


lancepioch:
@DaneEveritt Do we want CF specific instructions?

DaneEveritt:
I think this might make an already confusing subject for many people that much more confusing. Perhaps this would fit better in the community section @parkervcp made? Then it can specifically cover just CF setup for SSL?

parkervcp:
I agree that this is more of a "Community Section" change.

schrej:
@parkervcp we can probably close this one now?

PR #64, opened 2019-01-27
add minecraft server networks to community section

schrej:
This adds the guide on bungeecord networks from the deprecated readme.io community guides as it comes up rather frequently.
I've also added a section regarding ufw on ubuntu which causes issues quite often.

Do we want to have game specific sections in the community part of the docs or should we organise that differently?


DaneEveritt:
@parkervcp do you think it would be worthwhile to also extend this down the road to include the readme information you have in your eggs repo for specific games?

PR #66, opened 2019-01-31
Move docs over from guides site.

parkervcp:
Updating the Community Section About Page
Add Egg section from the guides site.
Start a Node config section with "Add a Node" first.

Added lots of images for community section.


schrej:
As you have grouped it as "game specific", which actually makes more sense, on the about page, could you rename the mc server network page from `minecraft/networks.md` to `games/minecraft.md`?
If you do, please also update the config to have it say "Game Specific" for the section and update the minecraft page so the first heading is "Minecraft", then the `[[toc]]` and then the rest, with every heading one level smaller.

PR #71, opened 2019-02-16
Change 0.7.11 to 0.7.12 (#2)

clrxbl:
Same as the other PR


DaneEveritt:
This doesn't actually need to be up to date, its just an example.

PR #73, opened 2019-03-05
Update to 0.7.13

MrKaKisen:
Replaces 0.7.12 in panel/upgrading to 0.7.13 which is the latest.


Sir3lit:
Shouldn't it just be 0.7.x like the daemons?

PR #77, opened 2019-03-27
Remove `or higher` from mysql

Sir3lit:
Mysql version 8.x isn't supported so `or higher` shouldn't be there, as it implies MySQL 8 is supported.


JanDevDE:
That's correct. For reference: https://mariadb.com/kb/en/library/mariadb-vs-mysql-compatibility/#replication-compatibility

PR #78, opened 2019-03-30
Restore the visitor's real IP address

Ethereal739:
Request Cloudflare to pass on the visitor's real IP address for session management


lancepioch:
This should be in a Cloudflare specific page

PR #79, opened 2019-03-30
Adding Fedora 29

Ethereal739:
Adding Fedora 29


DaneEveritt:
This should be in order of oldest -> newest, and if its following an existing one with matching OS name it doesn't need to declare it again.

PR #81, opened 2019-03-31
Use the official script instead of docker.io

Ethereal739:
Ubuntu 18.04 has been supported by docker for a while now, using the official script will give an up to date version of docker and is also more consistent with the official docs.


parkervcp:
LGTM

DaneEveritt:
```
Using these scripts is not recommended for production environments, and you should understand the potential risks before you use them
```

DaneEveritt:
If anything this documentation should be updated to follow what we have currently on the installation section which sends them to docker for OS specific steps.

PR #95, opened 2019-06-13
Remove redundant instructions to install certbot

Nuc1eoN:
Installing certbot at this point is redundant, because it is referred to later on anyways by linking to /tutorials/creating_ssl_certificates.html where the user can choose whether to use cerbot or a different tool


Nuc1eoN:
Should be updated for Ubuntu and Debian too, as well as #93 and maybe #94 ?

Issue #92, opened 2019-06-13
Wrong PHP install guide for CentOS


Nuc1eoN:
On the page https://pterodactyl.io/community/installation-guides/panel/centos7.html#php-7-2 it says to install php with the follwoing command:
`yum install -y php72u-php php72u-common php72u-fpm php72u-cli php72u-json php72u-mysqlnd php72u-mcrypt php72u-gd php72u-mbstring php72u-pdo php72u-zip php72u-bcmath php72u-dom php72u-opcache`

However there is no `php72u-mcrypt` package for php v7.2. Mcrypt has been removed from php 7.2.
The `php72u-php` package is also not available.
See https://dl.iuscommunity.org/pub/ius/stable/CentOS/7/x86_64/repoview/letter_p.group.html

Please fix the documentation if possible. How do I replace these packages?

lancepioch:
Yes, it needs to be updated. For now, you should be able to install PHP 7.3 by following: https://kifarunix.com/installing-php-7-3-3-on-centos-7-6/

You should be using Remi's repositories.

Nuc1eoN:
Thanks, got it to work as following if anybody wonders:
```
yum install epel-release yum-utils http://rpms.remirepo.net/enterprise/remi-release-7.rpm
yum-config-manager --disable remi-php54  --enable remi-php73
yum install php php-cli php-openssl php-gd php-mysql php-pdo php-mbstring php-tokenizer php-bcmath php-xml php-dom php-curl php-zip php-fpm php-pecl-zip
# Just to check:
php -v
```
As you see I also needed `php-pecl-zip` as composer complained it needs ext-zip while there is not such package. Weirdly this extension is not in the [required wiki packages ](https://pterodactyl.io/panel/getting_started.html#dependencies), but it seems to be needed at least for php v7.3 and composer.

DaneEveritt:
Again, PHP 7.3 is not an officially supported version at this time, which is why the documentation probably doesn't mention these things. Thanks for giving it a spin, but I'm going to close this PR.

Nuc1eoN:
@DaneEveritt This bug report was not about how to install php 7.3 but that the install commands for php 7.2 are wrong in the wiki.

DaneEveritt:
@parkervcp assigning this to you since I didn't write any of that community guide.

PR #98, opened 2019-06-21
Update PHP installation instructions

Nuc1eoN:
Ubuntu and Debian might also need to be updated

### DIFF
--- a/community/installation-guides/panel/centos7.md
+++ b/community/installation-guides/panel/centos7.md
@@ -47,7 +47,7 @@ yum-config-manager --disable remi-php54 --enable remi-php73
 yum update -y
 
 ## Install PHP 7.3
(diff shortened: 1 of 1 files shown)


Nuc1eoN:
Changes as suggested in #92

DaneEveritt:
7.2 remains the officially supported version, I will manually change this when I am ready to change the officially supported version and requirements.

Nuc1eoN:
The issue is that the current instructions are borked so any new user will not know what to do see #92

DaneEveritt:
Then the broken packages should be removed or addressed, I'm not comfortable just bumping the recommended version like that.

Nuc1eoN:
I just think it's better to have a working than a broken documentation but that is your decision.

PR #99, opened 2019-06-21
fpm extension is also needed

Nuc1eoN:


DaneEveritt:
This is incorrect if the user is running Apache under default conditions.

parkervcp:
Seeing as we favor nginx in our docs it may be worth it to call it out. Even if the install commands we offer install fpm already.

DaneEveritt:
Then it should probably have a little note in parenthesis mentioning it is only needed for nginx.

Nuc1eoN:
Added a clarification but feel free to alter it

lancepioch:
I'm wondering if it's worth it or not to just have a separate section for each Apache and Nginx.

PR #100, opened 2019-06-26
force-enable redis on start

Wqrld:
Redis doesn't auto start on some ubuntu versions to allow for configuration before redis is started. I've had multiple questions about this in the pterodactyl discord and believe it should be on by default do save people from issues where redis isnt online > their panel isnt working.


parkervcp:
This is actually not a bad idea to have.

ljavanbeek:
That's already added ( https://github.com/pterodactyl/documentation/blob/master/panel/getting_started.md#create-queue-worker , just above the pteroq enable command )

Wqrld:
okay my bad, thats new then

DaneEveritt:
You're enabling it before it is even installed though? Also it should go in the same place as the existing command and just modify that one to also boot it right then.

Wqrld:
see the new reference

PR #103, opened 2019-07-02
Link user to download files first [gh-pages]

JjTheHusky:
At the end of this community guide, it normally redirects the user to install the panel first, before downloading the files for it.

Changed from #installation to #download-files


DaneEveritt:
These are automatically compiled files and do not require a PR.

PR #104, opened 2019-07-07
Set SELinux permissions for redis

Nuc1eoN:
CentOS uses SELinux by default. So unless the user is using a non-default setup, this step is mandatory.

Seems to be an issue often experienced by CentOS users in the discord support channel

### DIFF
--- a/community/installation-guides/panel/centos7.md
+++ b/community/installation-guides/panel/centos7.md
@@ -125,7 +125,12 @@ Please check our [tutorial](/tutorials/creating_ssl_certificates.md) on generati
 <<< @/.snippets/webservers/nginx-centos.conf{5,11,26-27}
 
 ### Redis Setup
-The default Redis install is perfectly fine for the panel. If you have Redis already in use you may want to look into
+On systems using SELinux (which CentOS does by default) you will want to set the correct permissions:
+``` bash
+audit2allow -a -M redis_t
+semodule -i redis_t.pp
+```
+Apart from that the default Redis install is perfectly fine for the panel. If you have Redis already in use you may want to look into
 [running another Redis instance](https://community.pivotal.io/s/article/How-to-setup-and-run-multiple-Redis-server-instances-on-a-Linux-host).
 
 ## Installing the Panel


parkervcp:
Several vps providers like to disable selinux by default. That is why the steps are in the FAQ repo.

Nuc1eoN:
It is your decision I guess but I think an information like this should be included in the main instructions. If vps providers provide non-standard CentOS installs I don't see why that should be an issue for the CentOS guide.

Also that's why I chose the wording "On systems using SELinux..." so that there is no confusing. Another possibility would be to just add a `Note` paragraph pointing the user to https://pterodactyl.io/panel/troubleshooting.html#selinux-issues or similar.

I have had this issue myself and some people in the Discord support channel whom I helped out so I thought it might be helpful.

parkervcp:
audit2allow only works after there has been an error as far as I know.
From this [selinux](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/security-enhanced_linux/sect-security-enhanced_linux-fixing_problems-allowing_access_audit2allow) page directly

`generate SELinux policy allow rules from logs of denied operations"`

So running that before you ever get a denial would mean it's useless.

lancepioch:
Is there a way to allow it before it gets denied? If so, then we can put those instructions in and call it a day.

Nuc1eoN:
Ok I have come up with something like this, it might be positioned under "Redis Setup"
````
### Setting SELinux permissions
::: tip Note
While CentOS uses SELinux by default, some hosting providers tend to disable it in their CentOS images. In this case, or if you do not plan to use SELinux, you can skip the steps below.
:::
First check if SELinux is enabled on your system:
```bash
getenforce
```
If the output says `Enforcing` you can be sure that SELinux is enabled.
````
Feel free to modify it, it's just an idea. The commands to actually configure SELinux are missing, I did not know that the command I had included earlier only work after a denial, as @parkervcp has pointed out.

parkervcp:
I will work on this to have the rules set as part of the install.

parkervcp:
Closing this out in favor of #115

PR #105, opened 2019-07-08
Fix

lazarsn:
For to fix Access denied and 500 error after Access denied


schrej:
This is a bad pull request, sorry.
First of all: Use proper commit messages and a proper title for your pull request. "Fix" is not at all sufficient. Provide a brief description of what you changed so it's easier to see in the commit history. 
Also, providing random copy and paste commands in the documentation without any explanation whatsoever is not what we're trying to do. People should at least learn a little bit about those commands, if they don't know them already, so rather explain what they are doing exactly. Also explain what's causing the error. A http 500 error can literally be anything, you'd have to check the logs to see what's actually causing the issues.

As this seems to be SE Linux related, I'm closing this in favour of the much better PR #104

PR #111, opened 2019-08-06
Add phpMyAdmin Setup Guide

DomiiBunn:


DaneEveritt:
Thanks for taking the time to PR to our community, however I am not interested in having a community doc that is having people edit core files in the Panel. This will break each time the user upgrades, and will cause a flood of support requests into the Discord.

I am also on the fence about this guide in general, since installing PHPMA has really nothing to do with Pterodactyl, it just happens to be software people use along side it.

DomiiBunn:
I understand. I was whoever asked by a few people on how to add the menu item to the panel. I can remove the part. Or maybe adding "This edit is done at your own risk"?

lancepioch:
Thank you for your contribution, however we won't be merging this in due to multiple aforementioned issues currently.

Feel free to check with us in Discord in the future to be sure.

PR #115, opened 2019-08-17
Update/centos panel guide

parkervcp:
This is to update the centos guide and should also close #104 as it is fixing the redis perms


community/installation-guides/panel/centos7.md
```diff
@@ -35,7 +46,7 @@ systemctl start mariadb
 systemctl enable mariadb
 ```
 
-### PHP 7.2
+### PHP 7.3
 We recommend the ius repo to get the latest php packages.
```
Nuc1eoN: this should say remi

PR #119, opened 2019-08-21
Corrected the footer year

DomiiBunn:


schrej:
I think that doesn't really change the footer year on the frontpage. Not sure if it helps at all. I've opened another PR for the frontpage: #120

DomiiBunn:
I belie both of them are responsible for the footer :/ Tho i think the README takes priority with vuepress. Not sure tho

schrej:
That would be true if we'd actually use the page frontmatter on the Home.vue page. Maybe we should change that. Feel free to do so in this PR but please test if it works properly.

DomiiBunn:
Ima joink the reppo to my dev server give me a few

DomiiBunn:
you know where are these from?
```vue
    postcss: {
        plugins: [
            require('postcss-import'),
            require('tailwindcss')('./tailwind.js'),
            require('precss'),
            require('autoprefixer'),
            require('cssnano'),
        ]
    }```
I can't get vuepress to run cuz it can't find the modules :/

schrej:
Should be npm, run `npm install` or `yarn` if you have it

PR #120, opened 2019-08-22
Update footer year to 2019 on frontpage

schrej:


DomiiBunn:
I did approve it.....Don't understand GitHub at all. GitLab is way easier

schrej:
Well, you need to have write access in order for your approval to count. I think it's similar on GitLab.

DomiiBunn:
Well I checked it. Both of them should be changed whoever the one you updated is the one showing on the website.

PR #129, opened 2019-09-04
Add info on artisan commands.

DomiiBunn:


lancepioch:
artasin_commands.md

DomiiBunn:
Errr...Yes?..No? i mean that is the name of the file? But huh?

MatthewSH:
### `artasin_commands.md`

DomiiBunn:
Ah yes. i can not spell. xD Fixing it rn

lancepioch:
So here's some of my honest feedback that I'd like to share.

I really like the common errors section. I like it so much that I think we should have a dedicated page to it. Personally I'm imagining a page much less verbose than the troubleshooting.html page. Almost like a FAQ kind of setup.

Second, I think it's fine to have a list of our Artisan commands and what their parameters are. However, I believe that **either** for each of the individual commands their name is descriptive enough OR it should be changed so that it is.

I'm simply finding a lot of it to be redundant, for example:

Line 1: Server Rebuild
Line 2: php artisan p:server:rebuild <serverid>
Line 3: Rebuilds the specified server

To me, that says the exact same thing 3 times in a row. I think the command is descriptive enough and that if another command isn't, then it should be renamed to be.

DomiiBunn:
I was going for a guide where we can link to in support cases since many q are answered with "Run this command" etc. I did include the rest as a why not whoever i guess i didn't go into detail on those. I mean yeah the common errors thing is rly helpful. I can work on a pr for it as people ask for support in discord whoever you need to expect 100s of spelling errors.
The commands i am refairing to are `php artisan p:environment:setup` `php artisan p:server:bulk-power <power action> {--arguments}` `php artisan p:environment:mail` and `php artisan p:environment:database` people seam to find most issues with these. I guess since there are a part of the setup maybe integrating parts of this into the setup to provide a better description might be a thing?

DomiiBunn:
> I was going for a guide where we can link to in support
Main reason why all the way at the bottom and out of the way

lancepioch:
The discord bot does an amazing job providing instant and accurate feedback for most common issues and Parker deserves a ton of credit for cutting back on the total amount of support necessary.

So for example if pick on this command:

    php artisan p:server:bulk-power <power action> {--arguments}

image

I'm simply unconvinced that this huge amount of text is better than something like:

    php artisan p:server:bulk-power <stop,start,kill,restart> {--servers=1,2,3} {--nodes=1,2,3}

The only thing possibly missing would be a key explaining <required> and {--optional}

If a person has to be told that a command called "bulk-power" along with "start" will start more than 1 server, then they aren't our target or supported audience for our software.

DomiiBunn:
I have to agree the bot does an amazing job at catching all the errors! I mean this doc was just a suggestion so if you deem it unworthy I don't mind

DomiiBunn:
Closed due to #133

PR #132, opened 2019-09-06
Add a tip on escaping passwords

DomiiBunn:


tutorials/mysql_setup.md
```diff
@@ -67,7 +67,7 @@ CREATE USER 'data'@'127.0.0.1' IDENTIFIED BY 'password';
 ```
 Then add privileges for the user.
 :::danger
-If your database is on a different host than the one where your panel(daemon is installed, make sure to use `%` instead of `127.0.0.1` below
+If your database is on a different host than the one where your panel(daemon is installed, make sure to use `%` instead of `127.0.0.1` below. If your password contains special characters make sure to escape them
```
JanDevDE:
```suggestion
If your database is on a different host than the one where your panel(daemon is installed, make sure to use `%` instead of `127.0.0.1` below. If your password contains special characters, make sure to escape them.
```

lancepioch:
> ```[PDOException] SQLSTATE[HY000] [2006] MySQL server has gone away```
> The database server has closed the connection during a query. This would usually be caused by the SQL server being on the same server as the panel. Best way to resolve this issue is to set the IP to `127.0.0.1`

I don't believe this solution is correct, but I'm not 100% sure. Can I get a double check @arcdigital @MatthewSH ?

arcdigital:
This explanation isn't right. Usually this is some kind of configuration issue (often times with the socket - localhost uses a socket, 127.0.0.1 is over tcp).

DomiiBunn:
True. Whoever that was the fix for 3 cases on the support discord xD

MatthewSH:
That explanation feels very weird. I’d like to see the specific cases to have more context, but in general I think error documentation should fallback onto the software documentation. I’ll have to do more specific research on this topic but just a quick search led to MySQL docs regarding the issue: https://dev.mysql.com/doc/refman/8.0/en/gone-away.html

Obviously there’s more issues that can cause that specific error, but the way it’s described in the PR feels very strange and misleading. 

I just looked through a couple of the messages and it seems as though a common pattern is people creating database hosts and encountering this error, which in that case changing to 127.0.0.1 is not the solution, IICR, because it doesn’t display the correct credentials to the users of that host. While it does work, because they are hosted on the same machine, it shows that the end user has not properly configured MySQL to bind on all interfaces. 

There’s multiple reasons that could cause this error and I think context needs to be shown because there is not just a single solution for this.

DomiiBunn:
The [only one](https://discordapp.com/channels/122900397965705216/122903114423599104/618776260621107201) I can find but they deleted their messages :/

PR #133, opened 2019-09-08
Adding advanced usage section (w/ Artisan CLI)

MatthewSH:
While I like the idea of documenting functions from the Artisan CLI (first proposed in PR #129), there's things we have to be careful of and things **I** believe that were done incorrectly:

- I don't believe this should be classified as "Project Information" as most users will never need Artisan past the install process. The only time users who end up using Artisan are developers or advanced users and thus I think we should target those users. We can also expand this section and possibly move things like custom theming and addon creation to it in the future, if needed.

- Repetitive information. Since our target audience is advanced users, I don't believe we need to explain every little thing or simple concepts. Although I do add a few tips at the top to remind people at the beginning about a couple things, because we all can forget sometimes!

- Documenting commands that are used for installation reasons only. I found this kind of bugging me honestly. If a command is only being used once, during the initial setup...there's already enough information on it in the installation documentation. If there needs to be more information provided, then people can refer to the help command documented at the top of this Artisan documentation.

- Not grouping like commands. This cleans up the sidebar and provides an easier way for users to quickly find what they need.

---- 
A couple questions.

1) In the user management section you see this:
Create User
Should I leave it with these line-breaks or squish it all to one line? I didn't like how long it was on one line, that's why I did it like this. Thoughts?

2) Server rebuilding....I saw a couple different things. In the command itself it just pushes an update to the daemon and doesn't trigger the rebuild endpoint. Is the rebuild command on the CLI just for updating the daemons knowledge of the information or am I misunderstanding the code there?


DomiiBunn:
Yeah your implementation is way better than mine #129

MatthewSH:
@lancepioch @arcdigital could I get some second eyes on this to ensure I'm not missing any information or explaining things incorrectly?

DomiiBunn:
As far as i can see it's perfect!

.vuepress/config.js
```diff
@@ -100,6 +100,13 @@ module.exports = {
                         '/community/tutorials/custom_theme_setup.md',
                     ],
                 },
+                {
+                  title: 'Advanced Usage',
+                  collapsable: false,
+                  children: [
+                    '/community/advanced_usage/artisan.md',
```
DaneEveritt: This should just be under tutorials. There are a needless amount of sidebar sections. Can you also move node configuration to the tutorials section while you're at it?

community/advanced_usage/artisan.md
```diff
@@ -0,0 +1,153 @@
+# Artisan CLI
+
+::: tip
```
DaneEveritt: This should just be part of the intro paragraph, not a tip.

community/advanced_usage/artisan.md
```diff
@@ -0,0 +1,153 @@
...
+Artisan is a file located within your Pterodactyl installation directory. If you followed the official guide, it should be `/var/www/pterodactyl`.
+:::
+
+::: tip
```
DaneEveritt: Put this right below the help command, since thats the first place it gets used.

community/advanced_usage/artisan.md
```diff
@@ -0,0 +1,153 @@
...
+                        {--name-first=My}
+                        {--name-last=Name}
+                        {--password=supersecret}
+                        {--admin=1|0}
+                        {--no-password}
+```
+
+If you don't include the optional arguments, it will instead use an interactive prompt to guide you through the user creation process.
+
+### Delete/Remove User
```
DaneEveritt: This isn't `Delete/Remove`, its just `Delete`

community/advanced_usage/artisan.md
```diff
@@ -0,0 +1,153 @@
...
+
+To get information regarding a specific command you can do so by running:
+
+```bash
+php artisan help <command>
+```
+
+## User Management
+
+### Create/Make User
```
DaneEveritt: This isn't `Create/Make`, its just `Create`

community/advanced_usage/artisan.md
```diff
@@ -0,0 +1,153 @@
...
+```bash
+php artisan p:user:delete {--user=username/email/UUID}
+```
+
+If you don't include the optional argument, it will instead use an interactive prompt to ask you for the information on the user to remove.
+
+### Disable 2FA
+
+::: warning
+Disabling 2-factor authentication should only be used as a last restort for user recovery. **Please use this with caution**
```
DaneEveritt: Missing a period.

community/advanced_usage/artisan.md
```diff
@@ -0,0 +1,153 @@
...
+:::
+
+```bash
+php artisan p:user:disable2fa {--email=user@example.com}
+```
+
+If you don't include the optional argument, it will instead use an interactive prompt to ask you for the email address associated with the user.
```
DaneEveritt: Since this applies to all user commands, maybe just mention that at the start of the section. The user creation part will also prompt only for the things not passed through, so you can do a combo of args and interaction.

community/advanced_usage/artisan.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Server/Node Management
+
+### Create/Make Location
```
DaneEveritt: This is just create location.

community/advanced_usage/artisan.md
```diff
@@ -0,0 +1,153 @@
...
+Disabling 2-factor authentication should only be used as a last restort for user recovery. **Please use this with caution**
+:::
+
+```bash
+php artisan p:user:disable2fa {--email=user@example.com}
+```
+
+If you don't include the optional argument, it will instead use an interactive prompt to ask you for the email address associated with the user.
+
+## Server/Node Management
```
DaneEveritt: Server & Node Management

community/advanced_usage/artisan.md
```diff
@@ -0,0 +1,153 @@
...
+
+Displays a variety of panel information which can be used to check the configuation of this like database and email.
+
+### Cleanup Service Backups
+
+```bash
+php artisan p:maintenance:clean-service-backups
+```
+
+When modifying services, backup files are created. This command cleans up these backups in the storage directory.
```
DaneEveritt: This runs on a cron, users rarely need to run this.

community/advanced_usage/artisan.md
```diff
@@ -0,0 +1,153 @@
...
+
+### Cleanup API Keys
+
+```bash
+php artisan p:migration:clean-orphaned-keys
```
DaneEveritt: This was a one time command for the 0.7X migration, don't need it here.

.vuepress/config.js
MatthewSH: If you're talking about needless amount, shall I go ahead and move the creating and egg and docker image to tutorials as well? @DaneEveritt

community/advanced_usage/artisan.md
MatthewSH: I was kind of curious about this one when I was looking through the code, thanks! I'll remove it.

.vuepress/config.js
DaneEveritt: Nah, those kind of just fit where they are for now.

PR #135, opened 2019-09-11
Important! Fix mysql issue

DomiiBunn:
Fixed where it was suggesting to add a user to 127.0.0.1 user group and grant perms on a % user group user. This will cause a user not found error

### DIFF
--- a/tutorials/mysql_setup.md
+++ b/tutorials/mysql_setup.md
@@ -63,7 +63,7 @@ USE mysql;
 ```
 To then create a user run.<span style="color:red"> Change `data` and `password` to keep your database safe! </span>
 ```sql
-CREATE USER 'data'@'127.0.0.1' IDENTIFIED BY 'password';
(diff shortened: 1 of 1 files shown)


schrej:
There is a fat red warning below the SQL statement indicating to replace `127.0.0.1` with `%` if the database host is different than the panel host. If it's the same host, the current way is more secure.

schrej:
If you want to fix it, do it [here](https://github.com/pterodactyl/documentation/blame/master/tutorials/mysql_setup.md#L73).

DomiiBunn:
I am aware whoever. If you go through it one by one you'll find that you get an error of the user not existing.... I wouldn't add a comit like this to my own pr if I didn't find the issue....... Also since pteordactyl network isn't 127.0.0.1 this is an issue

DomiiBunn:
Or maybe i'm just dumb. both can be the case

schrej:
The created user is for the panel so it can connect to the DB in order to create and delete databases.

PR #137, opened 2019-09-13
Corrected the OS it's tailored for

YeapGuy:
Changed Debian 9 into Debian 10 since this is a Debian 10 guide.


parkervcp:
I apparently missed a copy/replace on the way.

PR #148, opened 2019-10-04
add sql warning for node db

parkervcp:
This is to add a warning about this not being how to configure a server to connect to a DB.

People probably won't read this either way.

### DIFF
--- a/tutorials/mysql_setup.md
+++ b/tutorials/mysql_setup.md
@@ -56,7 +56,7 @@ This section covers creating a MySQL user that has permission to create and modi
 :::
 
 ::: danger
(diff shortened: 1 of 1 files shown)


DaneEveritt:
What does this even mean? I wrote most of this page and this callout is not clicking at all.

parkervcp:
It's specifically calling out the setting up a db for a node section.

apparently people are saying it makes them think they can set their game servers to 127.0.0.1

DaneEveritt:
This now makes even less sense.

DaneEveritt:
Honestly I don't even see how this would do anything, people ignore all of the giant red boxes anyways. If they don't know they shouldn't use 127.0.0.1 this isn't going to change anything.

parkervcp:
I get that it's not being read in the first place by many, but I am sure this will at least catch some people.

PR #149, opened 2019-10-05
Fix DB host

lazarsn:


DomiiBunn:
Errr...what?

schrej:
The description field is there for a reason. The creation process is not broken currently, so if you want to change stuff, please explain why.
I also don't see a reason to shorten the user creation for mysql (if that's even correct that way).

PR #150, opened 2019-10-08

Sir3lit:
They really should be warnings.


DomiiBunn:
I actually agree...Maybe not warnings but something with a tab. also if u do 
::: warning <title> it'll make it look nicer no <>

PR #153, opened 2019-10-27
Add setup instruction for enabling Cloudflare on daemon

iamkubi:


parkervcp:
Not only are you (@iamkubi ) wrong about the daemon using cloudflare in front of the daemon requires a paid version of cloudflare because it uses websockets.

Not only can you have https over 8080 but we regularly do. it still serves the ssl certs correctly.

PR #154, opened 2019-10-27
Update php repo

SagnikSRHUSE:
Changed php repo's distro from stretch to buster. Currently, php7.x-curl's installation breaks as it's trying to find a version of libcurl that was included in stretch (Debian 9).


parkervcp:
Calling #157 integrates this and other changes suggested by just removing the repo all together. Closing this out for that one.

PR #157, opened 2019-11-01
update debian 10 install guide

parkervcp:
supersedes #151 #154 and #156

No extra repos are needed thus removing them
Changes the Debian 9 to Debian 10


devil38:
NodeJS 10 included
No Extra repos needed

PR #159, opened 2019-11-03
Update troubleshooting.md

SkitDev:
Updated one command in the file


parkervcp:
This brings the command in line with out support bot as they should be.

DaneEveritt:
@parkervcp can we also get the bot to use this command with the grep, would help people maybe better self-resolve issues since they won't get a long, often-confusing-unless-you're-used-to-it stacktrace.

parkervcp:
Done

Issue #158, opened 2019-11-01
PHP 7.3 in 'Getting Started' guide


vilhelmprytz:
I noticed PHP 7.2 was listed as dependency on the official 'Getting Started' guide ([Getting Started](https://pterodactyl.io/panel/getting_started.html#dependencies)). Should this be replaced with PHP 7.3 instead as PHP 7.2 is soon reaching end of 'active support' (see [supported versions](https://www.php.net/supported-versions.php)).

Cheers

Sir3lit:
Should be noted that security releases are supported until 30 Nov 2020

ljavanbeek:
PHP 7.2 was already replaced by PHP 7.3 in some of the community guides (CentOS 7 and the Debian guides), it just seems logical to replace it everywhere on the docs, right?

DaneEveritt:
It does, and likely will happen with the official 0.8 release since that is a great time to roll back through the documentation and get everything squared away. There is indeed nothing stopping people from using 7.3 though, and if someone would like to PR the necessary changes I'm happy to merge them early.

VilhelmPrytz:
> It does, and likely will happen with the official 0.8 release since that is a great time to roll back through the documentation and get everything squared away. There is indeed nothing stopping people from using 7.3 though, and if someone would like to PR the necessary changes I'm happy to merge them early.

I have created a PR for this https://github.com/pterodactyl/documentation/pull/160

(also accidentally closed and reopened the issue)

PR #160, opened 2019-11-04
Update PHP 7.2 to 7.3 in panel Getting Started guide

vilhelmprytz:
See https://github.com/pterodactyl/documentation/issues/158 for reference.


TrixterTheTux:
You'd also want to update the php-fpm socket path in the vhost config as otherwise people will get bad gateway.

VilhelmPrytz:
@TrixterTheTux Thanks for pointing out. I've added a new commit.

I replace 7.2 with 7.3 in the nginx configurations. A file named nginx-php7.3.conf already existed, so I deleted that one and updated all references for that file to point to the default file instead (I checked with diff before, it had not been modified in any other way).

Should be good now!

Cheers

VilhelmPrytz:
Any updates on this?

PR #162, opened 2019-11-05
Added Host name Chancing

Joshuaharry120:


parkervcp:
I would say this goes under tutorials and not panel.

DomiiBunn:
And you need to add it to the sidebar https://github.com/pterodactyl/documentation/blob/master/.vuepress/config.js

DaneEveritt:
This documentation does not meet the standards of the rest of the project. While I appreciate the attempt, this is a low quality document that does not sufficiently explain _why_ this documentation would be used, what the end result is, or walk users through any part of the process. It is effectively a list of commands, and many of them do not need to be run to adjust a domain in the system.

PR #163, opened 2019-11-13
Remove grant option for panel db

DomiiBunn:
As title

### DIFF
--- a/tutorials/mysql_setup.md
+++ b/tutorials/mysql_setup.md
@@ -46,7 +46,7 @@ include the `WITH GRANT OPTION` (which we are doing here). If you won't be using
 you can remove that.
 
 ``` sql
-GRANT ALL PRIVILEGES ON panel.* TO 'pterodactyl'@'127.0.0.1' WITH GRANT OPTION;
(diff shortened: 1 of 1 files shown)


DaneEveritt:
If you read the sorrounding documentation you'll see that this is important in context. In addition, your PR includes no relevant information as to why you're making this change, or what the benefits/fixes as a result of it are.

DomiiBunn:
Isn't the `GRANT OPTION` useless for just the panel as the next part of the guide https://pterodactyl.io/tutorials/mysql_setup.html#creating-a-database-host-for-nodes takes care of the nodes/server databases and i don't believe that the panel needs grant option for any other reason. Furthermore my panel instance (3 of em) all run on a db user without the `GRANT OPTION` perm. Dono tho. I would love to know what it is used for so? Since it just looks like a bit of a security risk atm....

schrej:
Maybe actually _read_ the text above the code block you're changing.

DomiiBunn:
Oh. I am just blind in multiple ways :^)

PR #166, opened 2019-11-22
Change fastcgi_pass to PHP 7.3

YeapGuy:
Changed the value of fastcgi_pass to be compatible with PHP 7.3. In the [Debian 10 Community Guide](https://pterodactyl.io/community/installation-guides/panel/debian10.html), you install PHP 7.3 but the config file is for 7.2.


parkervcp:
First of all #160 Already does this.
Second this would also effect the main docs which is why it hasn't changed.

PR #170, opened 2019-12-16
Add Archlinux Installation Documentation

0pcom:
I have added PKGBUILDS for the pterodactyl panel and daemon to the archlinux user repos:
https://aur.archlinux.org/packages/pterodactyl-panel
https://aur.archlinux.org/packages/pterodactyl-daemon

These can be built into p installed with yay, which will automatically handle the dependancies.

Scripts and a systemd service which were detailed in the other documenation are included in the built package. These are noted in the installation guide.


parkervcp:
You didn't update the `.vuepress/config.js ` file to add your guides;

0pcom:
I added it, sorry

parkervcp:
Your daemon page still doesn't load.

0pcom:
Should be good now

parkervcp:
We are now at my other comments about just linking them to the points in the main install docs for configuring the panel and daemon.

It is already covered. There is no reason to cover it twice.

0pcom:
The point of that was that I added a script at the package level which uses dialog that gives a menu of the configuration options. I found it easier to follow, to preface the daemon configuration section with that, however I can link per your specifications.

parkervcp:
Considering this is for Arch I assume these users are more advanced and can figure a ton of this out on their own.

0pcom:
Should be good now

community/installation-guides/panel/archlinux.md
```diff
@@ -0,0 +1,146 @@
...
+```
+
+The default Redis install is perfectly fine for the panel. If you have Redis already in use you may want to look into
+[running another Redis instance](https://community.pivotal.io/s/article/How-to-setup-and-run-multiple-Redis-server-instances-on-a-Linux-host).
+
+### Additional Utilities
+
+#### Certbot
+```bash
+apt install -y certbot certbot-nginx
```
andreas-fahrecker: Why use apt install in arch, especially if you have installed yay?

arcdigital:
stale, closing

PR #173, opened 2019-12-23
set permissions for wings file

markd69:
added a line to set perms for the wings file, I was having troubles running it without setting the perms.


parkervcp:
When you run `go build` it builds the file as executable. This should be needed.

markd69:
> When you run `go build` it builds the file as executable. This shouldn't be needed.

it's not referenced to build the daemon in the documentation... either way isn't it just easier to use the pre-built one?

parkervcp:
I see we are pulling the compiled version of the daemon now. Then this looks good to me.

PR #176, opened 2019-12-26
Minor Updates.

Sir3lit:
Minor updates to the wings install guide.

Fixed wings service file, fixed compatibility table for distro to match panel one. Fixe download link


Celemo:
Maybe Daemon does not work on the same distros as Panel and compatibility table is not accurate.

parkervcp:
The daemon works on all of those distros just fine.

PR #180, opened 2020-03-11
Combine ssl certificate tutorials

lancepioch:


schrej:
I think as this is not directly related to pterodactyl it might be a better fit for the community guides section.
Also make sure to update the configuration files when deleting pages, as it will break the build otherwise.

DaneEveritt:
@schrej I'm fine keeping it where it is for now, its pretty important for getting things running and seems to have stemmed a huge influx of questions since we created it initially. But yeah, gonna need a passing build :)

lancepioch:
This is the current behavior when searching:
image

The sections should be consolidated on one of the pages.

PR #181, opened 2020-03-12
Update deaemon installing.md

kingdevnl:
Change the line of npm install --only=production
To start with sudo, Some people seem to be getting this error

npm ERR! mmmagic@0.5.0 install: node-gyp rebuild
npm ERR! Exit status 1

It seems that using sudo often fixes the problem


lancepioch:
It is recommended against to run this as root. If it fails for a user, they are having other more important issues that they need to take care of.

PR #184, opened 2020-03-18
Added flag to skip the audit check

Pixel-Edwin:


DaneEveritt:
Ah thank you, I assumed there might be one, just couldn't find it.

PR #185, opened 2020-03-19
add npm flags to keep it from erroring

parkervcp:
This adds the `--no-audit` and  `--unsafe-perm` flags so it actually installs properly as root.


parkervcp:
Was added to the upgrade docs but not the everyday install docs.

PR #192, opened 2020-04-13
Add versioning

schrej:
All links keep working, at least as long as the current version stays the same.

`/[panel,daemon,wings]/current/` will redirect to the current version. Same applies if the version is omitted.

This also adds Wings to the sidebar. For now it's below Daemon and doesn't have a current version, we should move it up once 1.0 is released.


schrej:
don't merge yet

schrej:
should be good now

PR #195, opened 2020-04-16
Update Supported Systems

Sykursen:


parkervcp:
We are about to go through a major version release along with a new Ubuntu LTS release. I would rather wait until then to update any of these.

PR #199, opened 2020-04-19
Remove Ubuntu 14 / Remove forever

Sir3lit:
Removed forever, 99% of users are using a distro with systemd. If they're not..... Then.... Yeh they should update their distro unless there's one out there that people use and require it. 

Removed ubuntu 14.04 as nobody in their right mind should be using it.


Sir3lit:
Also updated the 1.0 to include 20.04 for when it's officially released on the 23rd. I'll test it to make sure everything is functional. 
Removed 16.04 / 14.04 also

PR #204, opened 2020-04-23
Added a note to help users with permissions for entrypoint.sh

Jpac14:
I have added this note to the bottom of page to help user set the right permissions for their entrypoint.sh I had this problem, so I would just like to let other user know before they have to take to the Pterodactyl Discord Server. I just thought it would be a nice note. I don't mind if you change the way it is said or formatted, as long as it has the same meaning as what I wrote.

### DIFF
--- a/community/config/eggs/creating_a_custom_image.md
+++ b/community/config/eggs/creating_a_custom_image.md
@@ -145,3 +145,6 @@ java -Xms128M -Xmx1024M -jar server.jar
 
 ## Run the Command
 The last step is to run this modified startup command, which is done with the line `${MODIFIED_STARTUP}`.
+
+### Note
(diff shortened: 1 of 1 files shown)


parkervcp:
Nowhere in my images to I set the permissions on the entrypoint.

The command to run an entrypoint should be important.

`/bin/bash entrypoint.sh` would use bash to run the entrypoint.sh

Jpac14:
Yes correct, but I did have to `chmod +x entrypoint.sh` then docker build my image. Then it all seemed to work, I will find you the error that I recieved

Jpac14:
Here was the error:
```
rror: (HTTP code 400) unexpected - OCI runtime create failed: container_linux.go:349: starting container process caused "exec: \"/entrypoint.sh\": permission denied": unknown 
```

matthewpi:
I've also had to do this when building images for my own use, otherwise they fail to boot as @Jpac14 has explained.

PR #206, opened 2020-04-28
Add upgrading page for 1. 0 

synocx:


schrej:
This only moves the upgrade guide between 1.x versions in the correct place and doesn't try to add a guide on how to upgrade from 0.7 to 1.0.

PR #212, opened 2020-05-15
Fixed some discrepancies

JordanPlayz158:
There are some errors like the config file location referencing the wrong spot so I just thought I'd fix it to help clear confusion.
(Based on develop branch of wings)


Sir3lit:
The daemon will look for the config in a few spots and move it to its proper location in the coming releases

DaneEveritt:
Should only need `/etc/pterodactyl` since the daemon will create everything else.

arcdigital:
I think this is fixed now

PR #216, opened 2020-05-19
Add instructions for disabling recaptcha

iamkubi:


matthewpi:
https://pterodactyl.io/panel/0.7/configuration.html#disable-or-modify-recaptcha

iamkubi:
Yeah, I thought it was there somewhere but failed to find it.  It turns out the docs only show up if you search for "pterodactyl disable recaptcha" but not "pterodactyl disable captcha".

matthewpi:
Even if you wanted to add it to that page you should link to the other one instead of writing it in two different places.

schrej:
I think we should move it to troubleshooting.

iamkubi:
I wrote the config:cache part based on a Github issue from 2017.  It's not in the current configuration guide, so I suspect it's not required.

PR #217, opened 2020-05-22
Add links to community guides for supported os table, uniformity changes within community guides.

Sir3lit:
Added links to OS versions to show the community guide that's specific to that OS and version.
Community guides now state they're for v0.7, I'm sure they would work for v1.0 but they should be tested to make sure.

### DIFF
--- a/community/installation-guides/daemon/centos7.md
+++ b/community/installation-guides/daemon/centos7.md
@@ -1,6 +1,5 @@
 # CentOS 7
-In this guide we will install Pterodactyl's Daemon 0.6.X — including all of it's dependencies — and configure it
-to use a SSL connection.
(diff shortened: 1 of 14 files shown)


schrej:
Could you _please_ give your PRs descriptive names in the future?
It's really annoying to have PRs with such a generic name. It's not that bad in the docs repo, but I've also seen it in the other repos as well.

Sir3lit:
I mean prs have a description box? Do you not look over the files that people are changing?

Sir3lit:
😊

schrej:
I do, but right now I have about 200 GH notifications, and when I clean up once in a while it helps to see what an issue/pr is about.
It doesn't have to be too specific, but even something like "Improve community guides" is better than just "doc changes", because I already know you probably modified the docs when you PR to the documentation repository 😄

PR #220, opened 2020-06-11
fix wings install location

schrej:
Executable files don't belong into `/etc`.
I'd suggest `/usr/local/bin` instead.


DaneEveritt:
Can you also add a quick thing to the upgrade guide to make sure people change that?

PR #221, opened 2020-06-11
Add TLS v1.3 as SSL protocol and update SSL ciphers in nginx configurations

vilhelmprytz:
Supporting TLS v1.3 is good practice and is what certbot currently does. See https://github.com/certbot/certbot/blob/896c1e0b66817eff447b41f619772a6441b416df/certbot-nginx/certbot_nginx/_internal/tls_configs/options-ssl-nginx.conf

Also updated the `ssl_ciphers` to what certbot uses.


DaneEveritt:
Sure, you can just do that in this PR as well.

vilhelmprytz:
:+1: This is now done!

Sir3lit:
Wasn't the issue that nodejs didn't support 1.3?

matthewpi:
> Wasn't the issue that nodejs didn't support 1.3?

Doesn't matter if Node.js supports it or not, NGINX doesn't proxy anything to Node.js and it would most likely do so over http anyways.

PR #230, opened 2020-07-07
Variable <phpversion> for 'fastcgi_pass' in webserver config

Toxicterror:
I think this change will help people not running into problems when certain parts of the documentation are refering to php7.2 and others to php7.3 or php7.4.

### DIFF
--- a/.snippets/webservers/nginx-nossl.conf
+++ b/.snippets/webservers/nginx-nossl.conf
@@ -24,7 +24,7 @@ server {
 
     location ~ \.php$ {
         fastcgi_split_path_info ^(.+\.php)(/.+)$;
-        fastcgi_pass unix:/run/php/php7.2-fpm.sock;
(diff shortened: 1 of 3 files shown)


DaneEveritt:
Reverting this, its just caused more confusion and made a process people already mess up a bunch even more likely to be messed up.

Toxicterror:
Is it possible to just mention it in the text above the code snippet without actually changing the snippet? Maybe that will help to reduce the amount of people messing up by a few.

What do you think?

DaneEveritt:
I think given that all but one guide explicitly install 7.2, leaving it explicit in the example is fine. I don't like increasing the number of variables. Even if it makes sense to all of us, there are plenty of users that, no matter _how many_ warnings and red boxes we put, will not read them. I'd rather just completely eliminate a potential pitfall, even if its not technically our fault. :)

PR #235, opened 2020-07-15
Remove 'warning Audit Results' warning.

ThatProgrammerCL:
Due to the new command change 'npm install --only=production --no-audit --unsafe-perm' instead of 'npm install --only=production', and due to the ``--no-audit' flag, the warning isn't needed.


parkervcp:
I disagree with removing it because some people think they know better and will just run `npm install --only=production` anyways.

ThatProgrammerCL:
I agree with Parker than some people don't just copy and paste and use the old one, but in the end.. most people copy and paste so I'm not too worried about someone cutting bits off one command.

PR #247, opened 2020-08-21
Changing PHP 7.4 to 7.2

Brouskat:
Since the stable release do not support PHP 7.4, let's change that.

### DIFF
--- a/.snippets/webservers/nginx-php7.2.conf
+++ b/.snippets/webservers/nginx-php7.2.conf
@@ -1,0 +1,66 @@
+server_tokens off;
+
+server {
+    listen 80;
+    server_name <domain>;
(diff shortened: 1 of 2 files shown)


FoksVHox:
> Since the stable release do not support PHP 7.4, let's change that.

I believe the docs were updated to support 1.0 once it releases

Brskt:
> > Since the stable release do not support PHP 7.4, let's change that.
> 
> I believe the docs were updated to support 1.0 once it releases

I see what you are referring but the thing is 1.0 is still in beta and people who have Ubuntu 20.04 have problems, come to get help on the Discord server and it's appear that they use PHP 7.4 and got issues with.

PR #248, opened 2020-08-21
Update ubuntu2004.md

Brskt:
Added the PHP repositorie since PHP 7.2 is not by default in Ubuntu 20.04
Fixed the links not redirect to the right #


arcdigital:
this should be updated for 1.0

PR #250, opened 2020-08-27
Update Installing page of the Daemon

Irgendwer01:
Adding Create Docker Network for Daemon


parkervcp:
The daemon auto-creates the network it needs.

parkervcp:
Closing as this is also not up to the standard that docs should have.

PR #251, opened 2020-08-31
Update Documentation

DeJayDev:
Make usage of `find` command POSIX compliant, and fixed an error in this `tar` command.

These commands barely work for me in the first place, but no one could even run them with a malformed tar command.

Logs that guided this issue
```
$ rm -rf $(find app public resources -d | head -n -1 | grep -Fv "$(tar -tf panel.tar.gz)")
find: warning: the -d option is deprecated; please use -depth instead, because the latter is a POSIX-compliant feature.
find: ‘app’: No such file or directory
find: ‘public’: No such file or directory
find: ‘resources’: No such file or directory

$ tar --strip-components=1 -xzv panel.tar.gz && rm -rf panel.tar.gz
tar: Refusing to read archive contents from terminal (missing -f option?)
tar: Error is not recoverable: exiting now
```


DaneEveritt:
Thanks, there was certainly some blind copy & paste action going on here on my end. 😆

PR #253, opened 2020-09-01
add steps to remove pterosftp when migrating to wings

schrej:


wings/1.0/migrating.md
```diff
@@ -51,6 +51,21 @@ rm -rf /srv/daemon
...
+```bash
+# stop and disable the standalone sftp
+systemctl disable --now pterosftp
+
+# delete the systemd service
+rm /etc/systemd/system/pterosftp.service
+
+# delete the sftp-server binary
+rm /srv/daemon/sftp-server
+```
```
SippingRetros, line 67: By this point, wouldn't the Daemon files already be removed thus the SFTP-Server being removed with it?

schrej: Indeed.

PR #259, opened 2020-10-10
Remove the recursive flag

markd69:
Removing the recursive flag to prevent damaging results.


parkervcp:
While not technically needed it's also not breaking anything to keep this here.

PR #270, opened 2020-11-08
Update community guides (php and maradb) and Update the majority of the docs to say 1.1 or 1.X

TekExplorer:
Update community guides to use php 7.4 and centos7 to use mariadb 10.5


TekExplorer:
as you can tell i need to learn git properly

TekExplorer:
BAD

PR #271, opened 2020-11-09
Update all the things to 1.1 or 1.X

TekExplorer:
Except maybe better


schrej:
I don't really like this approach.

TekExplorer:
Idk, I had time to burn. If you have any better ways I should do things lmk. There's a reason I put extra effort to split it from my other pr

TekExplorer:
I am very new to git and I don't know how you guys prefer to do things, so if you want to cherry-pick changes or would rather close the pr all together I won't mind. Any and all feedback is appreciated. First time I use git from the command line instead of GitHub itself.

schrej:
I'm referring to the approach of versioning for the docs, not your Git skills, in case you are confused.
I would prefer to actually duplicate the documents (or maybe just reference the same files) for the different versions, so we have proper versioning. That would also allow to copy-paste follow the guide when trying to install a previous version. We of course don't need to do it for patch versions.

In general it is better to have individual PRs for individual contributions, the same also applies to commits. That makes reviewing PRs easier and allows finer grained choices about what to accept and what to reject without additional effort.
Separating switching versions to 1.x and changing the php version into two PRs is the way to go.

DaneEveritt:
For versioning, I think until we need specifically different versions just keeping it under the 1.X space is fine. That's how I've always done it in the past, even when a specific version has some specific change necessary.

I think creating a new version for every 1.X release is just going to be tedious as fuck to maintain and link to the correct docs for, when realistically the versions just bump because of API changes and nothing to do with install/upgrade steps.

Perhaps just a landing page for upgrading that tells you what the minimum version of Wings for each Panel version is?

DaneEveritt:
I don't like this either, and it changes all of the live URLs for documentation that are linked across the internet. It isn't important that everything always reference the latest version when doing notices. I'm not solving for every user being an idiot, I'm solving for the users that can put 2 and 2 together to understand what is being discussed when I reference `1.0` and such.

Also this will perpetually be an absolute pain in the ass to keep updated.

TekExplorer:
fair enough. changing the links probably doesnt make much sense.

PR #273, opened 2020-11-09
Add hint to update nginx conf when upgrading php

schrej:
Not sure how this looks, but we can change it again if it's unreadable 😄 

Thanks to `J0N3S#1337` for pointing this out.


gamer4life1:
Maybe you should add a hint that if for some reason you still have php 7.2 (let's say from the old panel) and update mariadb, then it will not work and it will give you errors when migrating/upgrading. I don't know if that happens, but it happened to me so I personally think it would be helpful.
[@schrej]

PR #274, opened 2020-11-10
Optimize some images

gamer4life1:
Hello! I forked this repo to look around and it seems that the images could be all optimized. Here they are (credits to ImgBot for doing it...)


schrej:
generally a good Idea, we should check if the quality is still good enough though

schrej (approved):
thanks!

PR #276, opened 2020-11-10
mysql_native_password and caching_sha2_password

dudeisbrendan03:
mysql_native_password -> Not registered as an authentication plugin when it used to be the default, so should only be specified when caching_sha2_password is available else you'll get an error stating that mysql_native_password does not exist. Although if you are using a later version of MariaDB or MySQL and can't use caching_sha2_password you should specify this else you'll get an authentication error
caching_sha2_password -> Only supported by PHP 7.4+

### DIFF
--- a/tutorials/mysql_setup.md
+++ b/tutorials/mysql_setup.md
@@ -25,7 +25,7 @@ to our database. You can also use `%` as a wildcard or enter a numeric IP. We wi
 to `somePassword`.
 
 #### Later versions of Maria and MySQL (usually on latest versions of Debian distros)
-If you are using PHP 7.4+ then you should use the method for older vserions, as the default authentication method is correct! Laravel and PHP 7.4 can use the new caching_sha2_password authentication method.
+If you are using PHP 7.4+ then you should use the method for older versions, as the default authentication method is correct! Laravel and PHP 7.4 can use the new caching_sha2_password authentication method.
 
 ``` sql
 USE mysql;
--- a/tutorials/mysql_setup.md
+++ b/tutorials/mysql_setup.md
@@ -24,11 +24,25 @@ Next, we will create a user called `pterodactyl` and allow logins from localhost
 to our database. You can also use `%` as a wildcard or enter a numeric IP. We will also set the account password
 to `somePassword`.
 
+#### Later versions of Maria and MySQL (usually on latest versions of Debian distros)
+If you are using PHP 7.4+ then you should use the method for older vserions, as the default authentication method is correct! Laravel and PHP 7.4 can use the new caching_sha2_password authentication method.
+
 ``` sql
 USE mysql;
 
 # Remember to change 'somePassword' below to be a unique password specific to this account.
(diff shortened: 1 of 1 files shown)


Software-Noob:
You have a typo "vserions", also I'd personally leave the explanation of what it is and instead just have two alternative commands. Most users reading the documentation are just looking into copy-paste and get very easily confused and also easily skip steps.

The issue with users using mysql_native_password and it not working for their version gets brought up multiple times per day, so it would be great to have this included.

dudeisbrendan03:
> You have a typo "vserions", also I'd personally leave the explanation of what it is and instead just have two alternative commands. Most users reading the documentation are just looking into copy-paste and get very easily confused and also easily skip steps.
> 
> The issue with users using mysql_native_password and it not working for their version gets brought up multiple times per day, so it would be great to have this included.

People with no experience shouldn't be using this anyway, an explanation and references to the reason for the change is good in my opinion, thanks for the feedback!

I see it a lot so that's why I added it, some people get authentication issues because they aren't using mysql_native_password and others because they are on the Discord

TekExplorer:
I would probably change every instance of `maria` to `MariaDB`

DaneEveritt:
So, this documentation is very confusing to me, which I think also means it will only further confuse people who don't actually understand any of this. It should be highly explicit and use specific version ranges to guide people to the specific commands or configurations they need to use.

Software-Noob:
The part should be removed completely as it's not relevant to newer versions of 1.x or just a separate simple warning set, especially since 0.7x support is no longer there.

dudeisbrendan03:
1.x actually uses a later version of Laravel but I'm not sure if the version they're using supports the new password hashing method, I'll have to chec[](url)

Software-Noob:
> 1.x actually uses a later version of Laravel but I'm not sure if the version they're using supports the new password hashing method, I'll have to check

The requirement for specifying the authentication plugin should only be relevant to versions that are below the requirements for Pterodactyl 1.x

dudeisbrendan03:
Better idea, scrap all this and change the requirements
https://pterodactyl.io/panel/1.0/getting_started.html#dependencies
PHP 7.3 does not support caching_sha2_password, PHP 7.4 does.
Make the minimum requirement PHP 7.4

Software-Noob:
> Better idea, scrap all this and change the requirements
> https://pterodactyl.io/panel/1.0/getting_started.html#dependencies
> PHP 7.3 does not support caching_sha2_password, PHP 7.4 does.
> Make the minimum requirement PHP 7.4

This will be the case in 1.3 and forward

dudeisbrendan03:
> > Better idea, scrap all this and change the requirements
> > https://pterodactyl.io/panel/1.0/getting_started.html#dependencies
> > PHP 7.3 does not support caching_sha2_password, PHP 7.4 does.
> > Make the minimum requirement PHP 7.4
> 
> This will be the case in 1.3 and forward

Awesome

DaneEveritt:
7.4 is now the minimum version, closing this since it isn't relevant anymore.

PR #279, opened 2020-11-14
Adding more details on how to configure nginx

SAMXPS:
Hi, I am adding these commits on the documentation to help clients that had never used nginx before.
I guess it might be useful.


Software-Noob:
A more user-friendly way would be to create a symlink as done in the official guide, you also have a couple of typos such as "bellow " and "on" instead of "in".

DaneEveritt:
Sorry, but this does not meet the standards of the rest of the documentation and can be inferred/figured out from following the core documentation which explains how to configure SSL.

PR #281, opened 2020-11-23
Remove WITH mysql_native_password

TiziFioretti:


Software-Noob:
This is a duplicate PR and it would break it for versions that require the usage of authentication plugin, instead, it should provide both options and describe it like done in the PR [#276](https://github.com/pterodactyl/documentation/pull/276)

DaneEveritt:
If you're not going to provide reasoning for this change I'm not going to even look at it.

PR #285, opened 2020-11-29
Add 1.0 to 0.6 upgrading.md

TekExplorer:
add 1.0 upgrades to the old daemon's upgrading page

make of it what you will.


TekExplorer:
corrected `current` badges

DaneEveritt (approved):

PR #290, opened 2020-12-09
added FollowSymLinks

YouHaveTrouble:
solves 404 issues


DaneEveritt:
Please include more detail here, I have no idea what this is fixing or what the "404 issues" are.

YouHaveTrouble:
oh, while first installing 1.0 I ran with an issue where apache would throw 404 page not found error, adding those lines fixed it

DaneEveritt:
This does not seem necessary for normal installs, I think you just hit some edge case.

PR #293, opened 2020-12-17
New way to use mysql_native_password

dennorske:
Newer versions of MariaDB (tested on `10.5.8-MariaDB-1:10.5.8+maria~focal`) seems to not understand the current syntax- I worked around the problem by following guide here: https://mariadb.com/kb/en/authentication-plugin-mysql_native_password/#creating-users

Can someone else confirm my findings before accepting the PR?

### DIFF
--- a/tutorials/mysql_setup.md
+++ b/tutorials/mysql_setup.md
@@ -28,7 +28,9 @@ to `somePassword`.
 USE mysql;
 
 # Remember to change 'somePassword' below to be a unique password specific to this account.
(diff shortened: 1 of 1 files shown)


Software-Noob:
That will break the documentation for everyone who does not require the mysql_native_password authentication plugin. Also, this is a duplicate of #276

DomiiBunn:
This was changed as the Newer version requires it...

DomiiBunn:
Digging deeper, PHP 7.4 with laravel does support the new auth method so this can indead be removed outright. If someone is installing the old panel thier issue. might add a tip for that?

DomiiBunn:
Something like this would seam better 
image

PR #294, opened 2020-12-17
Add links to the OS specific community guides

TomDotBat:
Add links to all Wings supported operating systems with community guides as a shortcut.
This was done previously in the old daemon docs.


DomiiBunn:
Most of them are outdated. And following the main guide is your best bet for any deb basted os

DaneEveritt:
Agree with @DomiiBunn, I'd rather people discover those on their own after trying to follow the standard documentation which is the only one I ever touch when making changes.

PR #296, opened 2020-12-23
Fix create user command

mk7a:
mysql syntax error caused, changed to command format from pterodactyluser command below


Software-Noob:
There are a couple of duplicate pull requests for this, also this change without any notice will break this for other versions

AreYouRlyScared:
It's required with mysql 8  and higher version of mariadb due to the new mysql password method thats not supported ny php yet.

PR #297, opened 2020-12-23
Fix composer command

mk7a:
piped command requires sudo


TekExplorer:
these guides assume you are running as root user, so `sudo` was not added to any of them. if you are *not* running as root, its generally expected that you should be able to figure out where to put `sudo`

sergiz:
Composer should never be run as an elevated user.
https://getcomposer.org/doc/faqs/how-to-install-untrusted-packages-safely.md

PR #300, opened 2021-01-01
Specify 0.7 composer version

TekExplorer:
Been seeing many `undefined index: name` errors that could be quickly solved by adding this. yes, 0.7 deprecated, but there are many that install it anyway. let's save the headache. (inspiration from a pin in the discord)

### DIFF
--- a/panel/0.7/getting_started.md
+++ b/panel/0.7/getting_started.md
@@ -65,7 +65,7 @@ Composer is a dependency manager for PHP that allows us to ship everything you'l
 need composer installed before continuing in this process.
 
 ``` bash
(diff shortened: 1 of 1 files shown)


schrej:
Can that also be 1.x do we need the specific version? Or does the script not allow to use wildcards?

TekExplorer:
dunno, i just know its needed for 0.7

schrej:
I meant whether we could specify a wildcard for the composer install script to not pin it to an exact version

TekExplorer:
Ah, yeah, no idea

TekExplorer:
i dont see a variable, but i do see a direct download link: https://getcomposer.org/composer-1.phar
not sure if this is helpful in any way, or if we *have* to use the installer.

TekExplorer:
nope! i got it!

PR #303, opened 2021-01-02
remove certbot ppa

TekExplorer:
the repository is not needed for 20.04 or 18.04 from what i see, and it doesnt work on 20.04 anyway.


DaneEveritt (approved):

TekExplorer:
@DaneEveritt looks like I need another approval due to the conflict and minor addition

schrej (approved):

PR #308, opened 2021-01-07
Show versions separately :O

TekExplorer:
Make it so that wings and panel show 1.2 in the sidebar without destroying every link out there :)
Also made the old daemon collapsable so to minimize its footprint without outright removing it

as a side note:
package-lock.json changed... am I meant to include that?
gitpod also generated a .gitpod.yml, wasn't sure if I should include that either, so I left them out for now. can add the changes if asked

### DIFF
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -146,6 +146,7 @@ module.exports = {
                     currentVersion: '1.0',
                     versions: [
                         {
+                            title: "1.2",
                             name: '1.0',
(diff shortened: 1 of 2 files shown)


DaneEveritt:
Please use single quotes to keep it consistent with the rest of the code.

TekExplorer:
shoot, my bad, i didnt even notice

schrej (approved):

PR #312, opened 2021-01-13
Update about.md

vejskvit:
I think license that could be changed


DaneEveritt:
Keeping the license year perfectly up to date is not a concern of mine.

PR #314, opened 2021-01-19
Clarify Setting up MySQL (not sure if this is right)

Heath123:
From my experience it seems like they have to match

### DIFF
--- a/tutorials/mysql_setup.md
+++ b/tutorials/mysql_setup.md
@@ -61,7 +61,7 @@ If your database is on a different host than the one where your Panel or Daemon 
 ```sql
 USE mysql;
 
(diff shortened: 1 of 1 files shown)


Software-Noob:
You are creating a new database host account that will have grant permission for creating new databases for game servers, the current tutorial is correct. It has nothing to do with the existing panel login details.

Heath123:
> 
> 
> You are creating a new database host account that will have grant permission for creating new databases for game servers, the current tutorial is correct. It has nothing to do with the existing panel login details.

Yes, I realised that this is wrong, I'm still a bit confused though

Heath123:
I realised that you put the login details in on the panel...

PR #315, opened 2021-01-22
Better terminology

JackW6809:
Cleared up some of the wording. Not sure if all the changes I made are correct however that is the way I got it to work on my install.

### DIFF
--- a/guides/mounts.md
+++ b/guides/mounts.md
@@ -28,8 +28,8 @@ Mounts can not be mounted at `/home/container` or any subdirectory of it, as mou
 3. Fill in the details as required.
    - **Name**: Name for your mount.
    - **Description**: Description for your mount.
-   - **Source**: The directory where files are stored on the node.
-   - **Target**: The directory where the mount will be placed inside of your server, can **not** be `/home/container`.
+   - **Source**: The directory where files are stored on the node. E.g. `/example`
+   - **Target**: The directory where the mount will be placed inside of your server. E.g. `/home/container/plugins` You can **not** use `/home/container` by itself.
    - **Read Only**: Whether the mount will be read-only for the servers using it.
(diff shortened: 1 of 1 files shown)


matthewpi:
You cannot set the mount target inside of `/home/container` anywhere.  So `/home/container`, `/home/container/folder`, etc are all not allowed.

JackW6809:
> You cannot set the mount target inside of `/home/container` anywhere. So `/home/container`, `/home/container/folder`, etc are all not allowed.

@matthewpi Ah okay, so now what am I mean to put for the target then? An example would be great as from reading the documentation it is unclear to me thank you.

Software-Noob:
> > You cannot set the mount target inside of `/home/container` anywhere. So `/home/container`, `/home/container/folder`, etc are all not allowed.
> 
> @matthewpi Ah okay, so now what am I mean to put for the target then? An example would be great as from reading the documentation it is unclear to me thank you.

There was supposed to be an example screenshot in the documentation, but it wasn't added to it before the PR was merged.

Here's a gmod example:

Source: /home/cstrike
Target: /cstrike
Read-only: True
User mountable: False

You also have to add it to your allowed_mount in the wings config.yml file.

JackW6809:
> > > You cannot set the mount target inside of `/home/container` anywhere. So `/home/container`, `/home/container/folder`, etc are all not allowed.
> > 
> > 
> > @matthewpi Ah okay, so now what am I mean to put for the target then? An example would be great as from reading the documentation it is unclear to me thank you.
> 
> There was supposed to be an example screenshot in the documentation, but it wasn't added to it before the PR was merged.
> 
> Here's a gmod example:
> 
> Source: /home/cstrike
> Target: /cstrike
> Read-only: True
> User mountable: False
> 
> You also have to add it to your allowed_mount in the wings config.yml file.

Ah right, that is great thank you!

PR #317, opened 2021-01-26
Update additional_configuration.md

Software-Noob:
Remove s3 accelerate and style_endpoint as they're already set to false by default.

Removed .env section and fixed MySQL query typo together with shrinking it into a one-liner.


DaneEveritt:
Just remove the whole section with the .env edits, thats a hack solution. The DB query is not dangerous, that just encourages people to be scared of the database for absolutely no reason.

Software-Noob:
> Just remove the whole section with the .env edits, thats a hack solution. The DB query is not dangerous, that just encourages people to be scared of the database for absolutely no reason.

Your request has been fulfilled

PR #318, opened 2021-01-26
Update link to valid parsers

niekcandaele:
Idk if there should be a link to the old daemon parsing as well

### DIFF
--- a/community/config/eggs/creating_a_custom_egg.md
+++ b/community/config/eggs/creating_a_custom_egg.md
@@ -70,7 +70,7 @@ block prior to booting the server to ensure all of the required settings are def
 ```
 
 In this example, we are telling the Daemon to read `server.properties` in `/home/container`. Within this block, we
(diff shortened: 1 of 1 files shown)


community/config/eggs/creating_a_custom_egg.md
```diff
@@ -70,7 +70,7 @@ block prior to booting the server to ensure all of the required settings are def
 ```
 
 In this example, we are telling the Daemon to read `server.properties` in `/home/container`. Within this block, we
-define a `parser`, in this case `properties` but the following are [valid parsers](https://github.com/Pterodactyl/Daemon/blob/develop/src/helpers/fileparser.js):
+define a `parser`, in this case `properties` but the following are [valid parsers](https://github.com/pterodactyl/wings/blob/develop/parser/parser.go):
```
AreYouRlyScared, line 73: We should direct them to the highlighted lines 

https://github.com/pterodactyl/wings/blob/develop/parser/parser.go#L25-L30

AreYouRlyScared (changes requested):

PR #320, opened 2021-01-31
Add tabs for webserver configuration so page isn't so long

AreYouRlyScared:
Adds Tabs to web-server config for 1.X
Also remove package-lock.json as we use yarn :+1:
Removed the table of contents as it's kinda useless with tabs.

As seen below


schrej:
Redesigned them. The tab border is wider than the content, so the content width is consistent.
image

DaneEveritt:
Does it look okay if we just do a border on top rather than a whole box?

AreYouRlyScared:
I tried to remove the border but for the life of me I can't figure it out... @schrej might be able to?

TekExplorer:
why not just put all 4 options under one set of tabs? should make it a bit cleaner.

AreYouRlyScared:
hUA5ptX


Yeh I'd say that looks better.

AreYouRlyScared:
I still feel the tabs should have a box around them, as users might not notice they're clickable tabs.

DaneEveritt:
Should make them look like the GitHub tabs from this issue view IMO, simple and clear.

AreYouRlyScared:
image

I feel this defines the tabs more and also removes the border.
@DaneEveritt opinions?

DaneEveritt:
@AreYouRlyScared looks fine, can you fix the conflict?

AreYouRlyScared:
Il lbe able to when I get home. Trying to fix conflicts on mobile in kinda cancer

TekExplorer:
minor thing, none of these work except when its in the selected tab. 
ie. if I select the `Apache Without SSL` tab but press the `NGINX With SSL` in the sidebar, it doesn't do anything.
image

DaneEveritt:
@AreYouRlyScared we can just remove those from the sidebar since there are tabs now.

schrej:
Just make the headline smaller, then they should vanish from the sidebar

PR #324, opened 2021-02-11
Add alternative command for creating a MySQL user

wevetments:
Show users the alternative command to create a user in MySQL if a syntax error is thrown for using "mysql_native_password", but retain the original command for users that have the module.


wevetments:
Duplicate of #276

PR #327, opened 2021-02-13
Add information about where to create a new node

Software-Noob:
Instructions on where to actually create a node and that the config file must be created, as some users tend to have issues and questions related to this topic

### DIFF
--- a/wings/1.0/installing.md
+++ b/wings/1.0/installing.md
@@ -106,10 +106,9 @@ set when creating the node.
 :::
 
 ## Configure
-Once you have installed Wings and required components, the next step is to create a node on your installed Panel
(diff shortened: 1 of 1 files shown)


DaneEveritt:
This should probably also encourage the auto-deploy command, no?

Software-Noob:
> This should probably also encourage the auto-deploy command, no?

Possibly, I didn't go towards adding too much information to keep it short and simple since a lot of text gets skipped by users. I've also noticed that users tend to have issues with the auto-deploy occasionally

DaneEveritt:
If you don't mind adding maybe a short block about it after or something? At least users who can benefit from it might see it. :)

DaneEveritt (approved):

PR #329, opened 2021-02-19
Update mysql_setup.md

danymo1221:
Update from default SQL change to creating a user now uses IDENTIFIED


DaneEveritt:
This code is correct.

PR #331, opened 2021-02-24
Fix "Only for 0.7 panel" warning in the "Create theme" section

sinmineryt:
The warning wasn't showing as a box in the documentation, it's just showing this:
:::warn The theming system has not been ported to 1.0+. This guide is only applicable to 0.7.X. :::

Here is a screenshot what it is showing:
What it was showing

Here is a screenshot what it should show:
What it should show

It should look like this so people don't try this on the panel v1.x and then complain about it not working in the discord.


sinmineryt:
Closing this because 0.7 is EOL

PR #332, opened 2021-02-25
Update MySql Setup Page to drop an old command

WaffleThief123:
Updated the MySql setup docs to change the user create command to drop the `mysql_native_password` command from the user create, as this creates many problems and is not supported on modern things.


matthewpi:
https://github.com/pterodactyl/documentation/pull/329

PR #333, opened 2021-02-26
Add default egg variable list

Software-Noob:
This would be useful for new users that are looking into creating their custom eggs, as only three variables are documented anywhere on the panel itself.

### DIFF
--- a/community/config/eggs/creating_a_custom_egg.md
+++ b/community/config/eggs/creating_a_custom_egg.md
@@ -201,3 +201,20 @@ regex as any letters or numbers (`\w\d`) including underscore (`_`), periods (`.
 They will then be visible when managing the startup for a server in both the Admin CP and on the Front-End.
 
 ![](../../../.vuepress/public/community/config/eggs/Pterodactyl_Create_New_Egg_Startup.png)
+
+## List of default variables
+
+The default variables are always accessible to all eggs and don't have to be created separately. They can be used in the egg startup, install script, or the configuration file parser.
+
+| Variable | Description | Example |
+|----------|-------------|---------|
+| TZ       | Time Zone set from panels `.env` |  `Etc/UTC` |
+| STARTUP  | Startup command of the egg | `./Process` |
+| SERVER_MEMORY | Memory available for the server in MB | `512` |
+| SERVER_IP | Default ip of the server | `127.0.0.1` |
+| SERVER_PORT | Primary Server Port | `27015` |
+| P_SERVER_LOCATION | Location of the server | `Example City` |
+| P_SERVER_UUID | UUID of the server | `539fdca8-4a08-4551-a8d2-8ee5475b50d9` |
+| P_SERVER_ALLOCATION_LIMIT | Limit of allocations allowed for the server | `0.000000` |
+| USER | User that executes the startup command in the server | `container` |
+| HOME | Home path of the container user | `/home/container` |
--- a/community/config/eggs/creating_a_custom_egg.md
+++ b/community/config/eggs/creating_a_custom_egg.md
@@ -208,13 +208,11 @@ The default variables are always accessible to all eggs and don't have to be cre
 
 | Variable | Description | Example |
 |----------|-------------|---------|
-| TZ       | Time Zone set from panels `.env` |  `Etc/UTC` |
-| STARTUP  | Startup command of the egg | `./Process` |
+| TZ       | Time Zone |  `Etc/UTC` |
+| STARTUP  | Startup command of the egg | `java -Xms128M -Xmx{{SERVER_MEMORY}}M -jar {{SERVER_JARFILE}}` |
 | SERVER_MEMORY | Memory available for the server in MB | `512` |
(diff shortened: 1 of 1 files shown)


community/config/eggs/creating_a_custom_egg.md
```diff
@@ -201,3 +201,20 @@ regex as any letters or numbers (`\w\d`) including underscore (`_`), periods (`.
...
 ![](../../../.vuepress/public/community/config/eggs/Pterodactyl_Create_New_Egg_Startup.png)
+
+## List of default variables
+
+The default variables are always accessible to all eggs and don't have to be created separately. They can be used in the egg startup, install script, or the configuration file parser.
+
+| Variable | Description | Example |
+|----------|-------------|---------|
+| TZ       | Time Zone set from panels `.env` |  `Etc/UTC` |
+| STARTUP  | Startup command of the egg | `./Process` |
```
DaneEveritt, line 212: Might be good to use an actual line from the vanilla MC server or something showing that the variables aren't replaced in this yet.

community/config/eggs/creating_a_custom_egg.md
```diff
@@ -201,3 +201,20 @@ regex as any letters or numbers (`\w\d`) including underscore (`_`), periods (`.
...
+| SERVER_MEMORY | Memory available for the server in MB | `512` |
+| SERVER_IP | Default ip of the server | `127.0.0.1` |
+| SERVER_PORT | Primary Server Port | `27015` |
+| P_SERVER_LOCATION | Location of the server | `Example City` |
+| P_SERVER_UUID | UUID of the server | `539fdca8-4a08-4551-a8d2-8ee5475b50d9` |
+| P_SERVER_ALLOCATION_LIMIT | Limit of allocations allowed for the server | `0.000000` |
```
DaneEveritt, line 218: This isn't a float.

community/config/eggs/creating_a_custom_egg.md
```diff
@@ -201,3 +201,20 @@ regex as any letters or numbers (`\w\d`) including underscore (`_`), periods (`.
...
 
 ![](../../../.vuepress/public/community/config/eggs/Pterodactyl_Create_New_Egg_Startup.png)
+
+## List of default variables
+
+The default variables are always accessible to all eggs and don't have to be created separately. They can be used in the egg startup, install script, or the configuration file parser.
+
+| Variable | Description | Example |
+|----------|-------------|---------|
+| TZ       | Time Zone set from panels `.env` |  `Etc/UTC` |
```
matthewpi, line 211: This is not pulled from the Panel's `.env`, it is pulled from the wings configuration and can be overridden by an egg variable.

community/config/eggs/creating_a_custom_egg.md
```diff
@@ -201,3 +201,20 @@ regex as any letters or numbers (`\w\d`) including underscore (`_`), periods (`.
...
+| STARTUP  | Startup command of the egg | `./Process` |
+| SERVER_MEMORY | Memory available for the server in MB | `512` |
+| SERVER_IP | Default ip of the server | `127.0.0.1` |
+| SERVER_PORT | Primary Server Port | `27015` |
+| P_SERVER_LOCATION | Location of the server | `Example City` |
+| P_SERVER_UUID | UUID of the server | `539fdca8-4a08-4551-a8d2-8ee5475b50d9` |
+| P_SERVER_ALLOCATION_LIMIT | Limit of allocations allowed for the server | `0.000000` |
+| USER | User that executes the startup command in the server | `container` |
```
matthewpi, line 219: `USER` and `HOME` may not always be set as they are specified in the `Dockerfile` of the image being used, no one should really need to use these environment variables anyways.

community/config/eggs/creating_a_custom_egg.md
Software-Noob: Whenever I call for the variable, it returns a float. 


echo "location limit is ${P_SERVER_ALLOCATION_LIMIT}"
location limit is 0.000000

community/config/eggs/creating_a_custom_egg.md
Software-Noob: They appear to be always defined in all official images at least. Do you think they should be removed from the list?

matthewpi: > They appear to be always defined in all official images at least. Do you think they should be removed from the list?

They aren't really important to have so I don't see why they should be there.  Also, this guide is on eggs specifically where people may be using custom images.  I'd just remove them

community/config/eggs/creating_a_custom_egg.md
matthewpi: > Whenever I call for the variable, it returns a float.
> 
> echo "location limit is ${P_SERVER_ALLOCATION_LIMIT}"
> location limit is 0.000000

Then that is a bug that should be fixed, that value should be an integer.

DaneEveritt (approved):

PR #334, opened 2021-02-26
Remove themes since they're no longer relevant

Software-Noob:
Removes guides regarding themes since 0.7 is EOL and they're not relevant to the current 1.x versions


DaneEveritt:
Please don't delete the actual file about the themes, only references to it (and the artisan commands are fine). I don't like when old documentation just disappears, it can be nice to reference down the road and people who genuinely need it can still find it with a bit of effort.

DaneEveritt:
Scratch that, this documentation isn't all that advanced in the first place.

schrej (approved):

PR #335, opened 2021-02-26
Added OVH disclaimer to installing.md

kytobias:
Just thought it would be good to add this kind of notice to configuring Wings as I spent the last few hours scratching my head on this until I realized that I couldn't actually uses Cloudflare's or Google's DNS servers with my host which caused errors like this: 

[https://cdn.discordapp.com/attachments/361241250915024906/814740776189165588/unknown.png](https://cdn.discordapp.com/attachments/361241250915024906/814740776189165588/unknown.png)

### DIFF
--- a/wings/1.0/installing.md
+++ b/wings/1.0/installing.md
@@ -112,6 +112,10 @@ After you have created a node, click on it and there will be a tab called Config
 
 Alternatively, you can click on the Generate Token button, copy the bash command and paste it into your terminal.
 
+::: warning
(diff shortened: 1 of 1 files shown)


iamkubi:
External DNS servers work fine on OVH.  You were most likely experiencing https://github.com/gliderlabs/docker-alpine/issues/334 which is a Docker bug and has nothing to do with your host.

kytobias:
Ah  I see, would it still count as a viable way of solving the bug for Pterodactyl specifically?

iamkubi:
It could, but it may not be entirely reliable.  From what I understand there are weird caching issues with the DNS proxy that Docker uses internally.  I do think that changing your DNS servers and restarting Docker is _usually_ enough to fix it, but I've also had people say that did not resolve it for them.

schrej:
Based on @iamkubi's comments I don't think this is useful to have, thanks though.

PR #337, opened 2021-03-02
Changed SQL command to create user.

sammilt:
`WITH mysql_native_password` is no longer required in MariaDB/MySQL to create the user and returns an error when executed.


Software-Noob:
Duplicate of #276

PR #339, opened 2021-03-05
1.3.0

matthewpi:
PHP 8 upgrade guide may still need some work but this should be everything we need for the 1.3.0 upgrade guide.


panel/1.0/getting_started.md
```diff
@@ -206,7 +210,7 @@ If you are not using `redis` for anything you should remove the `After=` line, o
 when the service starts.
 :::
 
-If you are are using redis for your system, you will want to make sure to enable that it will start on boot. You can do that by running the following command: 
+If you are are using redis for your system, you will want to make sure to enable that it will start on boot. You can do that by running the following command:
```
Software-Noob, line 213: While you're at it, maybe fix the double `are` that has been here for a while since making a PR for that one typo feels strange

PR #341, opened 2021-03-08
Update updating.md

Oberfail:
A couple of people didn't knew, they had to be in the correct directory to do php artisan down, so I suggest to add it.


dudeisbrendan03 (changes requested):
This logically makes more sense, but you've removed the explanation of what the command in "Download the update" is doing.

You should add back what piping the curl output from the remote into tar will do on the users machine.

Software-Noob:
> This logically makes more sense, but you've removed the explanation of what the command in "Download the update" is doing.
> 
> You should add back what piping the curl output from the remote into tar will do on the users machine.

I think you are misinterpreting the changes as that section was not removed and is still present

dudeisbrendan03 (approved):
@Software-Noob 
My bad, I'm blind apparently

PR #342, opened 2021-03-08
Change the base docs to 20.04

hwalker928:
Changed the table to show that 20.04 is the base OS that the documentation is written in.
Most people now are using the newer Ubuntu 20.04 OS, meaning the docs should also be where most people who are installing for the first time are using.


dudeisbrendan03:
On one side the documentation currently covers 20.04 and 18.04, there's nothing wrong with that, but on the other side 18.04 is 2 years old and has entered its first stage of it's EOL, although it still has maintenance updates.

I'd probably leave it as is.

PR #343, opened 2021-03-08
GitHub Workflows: Close stale issues automagically, never close a PR

dudeisbrendan03:
There are issues from mid-2020, close stale issues automatically with a workflow.


DaneEveritt:
I despise repositories that do this, there is nothing beneficial for an organization of this size to just close issues that are "stale". Tons of issues and PRs just sit because I don't have the time or motivation to deal with them yet.

Thanks, but there is no benefit from adding this IMO.

PR #344, opened 2021-03-09
Changed daemon references to wings

hwalker928:
There were still links leading to 404 pages (old daemon tutorials). I have changed these to the wings equivalent.


Software-Noob:
> are you aware that those links are also a 404...? the new links are .html now
> 
> ex: https://pterodactyl.io/community/installation-guides/panel/centos8.html

They're all markdown files that are converted into HTML. They'll work fine being referenced as .md files in the files, and that's how they're referenced everywhere in the documentation files

PR #346, opened 2021-03-12
Various updates

Boy132:
Add node:
- Updated some links and wordings
- Updated some images to use the new 1.x theme

Additional configuration:
- Added 2FA section

Panel/ Getting started:
- Fixed typo

PS: sorry for the "commit spam" :P


community/config/nodes/add_node.md
```diff
@@ -43,7 +43,7 @@ Don't forget to account for OS overhead and other software requirements on machi
 * **Daemon SFTP Port**: the port the Daemon sftp-server or standalone SFTP server should listen on.
 
 ## Install the Daemon
-At this point you'll need to have the Daemon installed on your machine. Check out the [documentation](/wings/installing.html)
+At this point you'll need to have the Daemon installed on your machine. Check out the [documentation](/wings/1.0/installing.html)
```
schrej, line 46: leaving out the 1.0 is actually correct here, as it'll (or at least should) redirect to the latest version.

schrej (changes requested):

community/config/nodes/add_node.md
Boy132: > leaving out the 1.0 is actually correct here, as it'll (or at least should) redirect to the latest version.

You're right! Thanks for pointing out. :)

schrej (approved):

Kaesewuerfel:
Very good work!!!!!! you are doing great!!!!!! look forward to the next generation!!!!!!!1 GREAT!!!!!

PR #353, opened 2021-03-25
Add Wings to the chronological next step

Software-Noob:
Some users appear to be missing that you actually need to install Wings and not just the Panel. This would make it a complete flow of Panel installation > webserver configuration > Wings installation

### DIFF
--- a/panel/1.0/webserver_configuration.md
+++ b/panel/1.0/webserver_configuration.md
@@ -86,4 +86,4 @@ systemctl restart apache2
 :::
 ::::
 
-#### Next Step: [Wings Installation](../../wings/./installing.md)
+#### Next Step: [Wings Installation](../../wings/installing.md)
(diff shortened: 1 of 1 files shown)


panel/1.0/webserver_configuration.md
```diff
@@ -85,3 +85,5 @@ systemctl restart apache2
 
 :::
 ::::
+
+#### Next Step: [Wings Installation](../../wings/./installing.md)
```
DaneEveritt, line 89: why the random dot?

DaneEveritt (changes requested):
see comment

panel/1.0/webserver_configuration.md
Software-Noob: Looks like my vscode auto-suggest filled that in, but it works even with the dot as I tested prior to the PR

Removed the random dot

DaneEveritt (approved):

PR #354, opened 2021-04-04
Added app_name

FoksVHox:
Documents pterodactyl/wings#88


TASelwyn:
Uhm, 
Shouldn't the default say "Pterodactyl" not "null" ?

FoksVHox:
> Uhm,
> Shouldn't the default say "Pterodactyl" not "null" ?

It should, my bad

schrej (approved):

PR #358, opened 2021-04-28
Update debian10.md to install php8.0

HiKE-yt:
Modified to install php8.0 and all it's dependencies in 8.0 (previous : 7.4)

### DIFF
--- a/community/installation-guides/panel/debian10.md
+++ b/community/installation-guides/panel/debian10.md
@@ -23,13 +23,13 @@ systemctl start mariadb
 systemctl enable mariadb
 ```
 
-### PHP 7.4
+### PHP 8.0
 ```bash
 ## Get apt updates
 apt update
 
 ## Install PHP 7.4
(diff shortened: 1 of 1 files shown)


community/installation-guides/panel/debian10.md
```diff
@@ -23,13 +23,13 @@ systemctl start mariadb
 systemctl enable mariadb
 ```
 
-### PHP 7.4
+### PHP 8.0
 ```bash
 ## Get apt updates
 apt update
 
 ## Install PHP 7.4
```
Software-Noob, line 31: This comment should be changed to PHP 8.0.

I am fairly sure Debian 10 will not have PHP 8.0 packages out of the box. Doesn't it require you to add a new source, for example, from sury? I think only difference between official documentation and Debian ones are the PPA repositories

Software-Noob, line 31:
```suggestion
## Install PHP 8.0
```

Software-Noob (approved):
The PHP source repository was added with #372. As such, this will now work as you intended.

PR #362, opened 2021-05-06
Fix url at the end of wings community guide

JohnB17:
All of the wings community guides don't have the correct link at the end.


DaneEveritt:
Why did you make them all link to a specific subsection?

JohnB17:
The original one did, but it was changed in the official wings documentation, also it makes sure they start at that section instead of going straight to the bottom (which it did for me)

Software-Noob (approved):

PR #364, opened 2021-05-16
Added restart delay for pteroq and wings service

Simonmicro:
There is a chance, that e.g. the database server of the system is not yet running after any reboot. Without the `RestartSec=` option systemd will restart the service with an delay of just `100ms` for `DefaultStartLimitBurst` in an interval of `DefaultStartLimitIntervalSec`. This may also fail as the database could take a little bit longer. In that case the startup of the service is aborted and the queue worker dies silently.

Using this option (set to `30s`) should slow down this behavior and also prevent systemd of giving the service up.

I've also extended the wings service, as it seems also to be affected slow database startups (when the respective host is simultaneously (re-)starting).

### DIFF
--- a/panel/1.0/getting_started.md
+++ b/panel/1.0/getting_started.md
@@ -196,6 +196,9 @@ User=www-data
 Group=www-data
 Restart=always
 ExecStart=/usr/bin/php /var/www/pterodactyl/artisan queue:work --queue=high,standard,low --sleep=3 --tries=3
+StartLimitInterval=180
+StartLimitBurst=30
+RestartSec=5s
 
 [Install]
 WantedBy=multi-user.target
--- a/wings/1.0/installing.md
+++ b/wings/1.0/installing.md
@@ -157,7 +157,9 @@ LimitNOFILE=4096
 PIDFile=/var/run/wings/daemon.pid
 ExecStart=/usr/local/bin/wings
 Restart=on-failure
-StartLimitInterval=600
+StartLimitInterval=180
+StartLimitBurst=30
+RestartSec=5s
 
 [Install]
 WantedBy=multi-user.target


Simonmicro:
**Bump**

Software-Noob:
Such a high delay would make sense for a system reboot, but it's likely too high for crashes?

Simonmicro:
But as Wings just crashes, when the api endpoint is unavailable, we have currently no way of identifying a crash by api failure or crash by some buggy code.

As I have not (yet) seen any crashes by bugs (but only dead services by the api failing to respond) I tend to just generally slow down the Wings restart. Especially as Systemd gives up any service restarting & crashing too fast, this would also prevent Wings from just dying if it is getting unstable for whatever reason.

Are there any alternative solutions, which you would prefer? I mean alternatively, we could just add a note below the service file, explaining the problem and the recommended solution.

Software-Noob:
30 seconds is a too long time for someone to wait to see if their Wings service begins to work, especially considering a lot of users don't even know how the service works. 

We have a service file with `StartLimitInterval=600` seconds and `StartLimitBurst` is not set thus the [default is 5](https://www.freedesktop.org/software/systemd/man/systemd-system.conf.html#DefaultStartLimitIntervalSec=). Because of this, the service is allowed 5 restarts within 600 seconds before entering the failed state forcing users to perform `systemctl reset-failed` to fix their wings.service even if nothing would stop it from running. I've seen this many times in Discord but didn't really think of it until now.

As Restart is not specified it defaults to `DefaultRestartSec` 100ms reaching the failed-state pretty much right away. I think we should set the following:
```
StartLimitInterval=50
StartLimitBurst=10
RestartSec=5s
```
Allowing us to restart ten times, every five seconds until entering the failed state. This would be a huge upgrade over the current five times every 100ms. It wouldn't cause too long wait times between crashes and also allows the service to restart itself after the system reboots.

This will only affect new installs or users that would manually update their service files.

Simonmicro:
> 30 seconds is a too long time for someone to wait to see if their Wings service begins to work

It is a _restart_ delay, therefore only relevant on the aforementioned system reboots, or more generally if the service got an _unclean_ exit result.

> It wouldn't cause too long wait times between crashes and also allows the service to restart itself after the system reboots.

Fair solution - I would still tend to retry the service for a long time (especially larger database servers like to take a while to start), so I would propose at least:
```systemd
StartLimitInterval=180
StartLimitBurst=30
RestartSec=5s
```

If this is fine by you, I'll update the PR accordingly.

> This will only affect new installs or users that would manually update their service files.

Indeed - but what else should we do?

Software-Noob:
> > 30 seconds is a too long time for someone to wait to see if their Wings service begins to work
>
> It is a _restart_ delay, therefore only relevant on the aforementioned system reboots, or more generally if the service got an _unclean_ exit result.

I am aware, the service can fail during runtime, and having a 30 second wait time is quite long considering we had 100ms before. A shorter restart time with higher bursts would be a better solution. 

> > This will only affect new installs or users that would manually update their service files.
> 
> Indeed - but what else should we do?

I am pointing this out for anyone else that will read this

---

> > It wouldn't cause too long wait times between crashes and also allows the service to restart itself after the system reboots.
> 
> Fair solution - I would still tend to retry the service for a longer time (especially larger database servers like to take a while to start), so I would propose at least:
> 
> ```
> StartLimitInterval=180
> StartLimitBurst=30
> RestartSec=5s
> ```
> 
> If this is fine by you, I'll update the PR accordingly.

cc @pterodactyl/documentation for this change to fix the service restarts?

Simonmicro:
@Software-Noob here you go - I hope this is fine now!
@pterodactyl/documentation Ping.

Software-Noob (approved):
LGTM

DaneEveritt:
I agree with these changes. 👍🏼

PR #365, opened 2021-05-18
Update updating.md

mtjohnson47:
Panel version 1.2.2 gives errors when trying to updating using php 8

### DIFF
--- a/panel/1.0/updating.md
+++ b/panel/1.0/updating.md
@@ -25,7 +25,7 @@ Composer v2.
 * Composer `2.X`
 
 Previous versions of Pterodactyl allowed for you to be running PHP 7.3. However, due to dependency updates and
-the fact that 7.3 is now EOL we have bumped the minimum acceptable PHP version to 7.4 — PHP 8 is recommended. You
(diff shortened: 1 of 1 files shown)


DaneEveritt:
I don't follow what you're fixing here, can you explain more please?

mtjohnson47:
On 1.2.2, putting the panel in maintenance mode gives errors if using PHP 8, and has to be done with PHP7.4.  So, only update to PHP8 is already on 1.3 or higher.

DaneEveritt:
We don't expect people to only update to a release that is no longer supported, this change is unnecessary.

mtjohnson47:
You misunderstand.  I'm not talking about updating to 1.2.2, I'm talking about updating FROM 1.2.2.

DaneEveritt:
Ah, I'm really not that worried about it seeing as that update has been out for months now. Not being able to put the panel into Maintenance mode isn't the end of the world, if anything someone can just move the down command above the dependency update which is probably more correct anyways.

The wording you added isn't quite intuitive (even to me) and just encourages people to not update all the way to 8, which is highly recommended to avoid having to repeat this process down the road.

PR #366, opened 2021-05-19
Added a warning about updating to PHP 8 on panel version 1.2.2 or older.

mtjohnson47:
On Panel 1.2.2 and older, putting the panel in maintenance mode gives errors.  I added a warning to not update to PHP 8 until after the panel is in maintenance mode.  I reworded my previous pull request at the maintainer's suggestion.


Software-Noob:
The more reasonable solution would be to move the artisan down command above the dependency install. There's no need to have all that repetitive text

PR #367, opened 2021-05-26
Add Docker to terminology

Shadowhackercz:
I think the Docker should be in the terminology too, so I added it, feel free to change


Software-Noob (approved):
LGTM, thank you for your contribution!

PR #368, opened 2021-06-01
Not access to Panel

Justman10000:
Without this, access to the panel will be denied!

### DIFF
--- a/.snippets/webservers/apache.conf
+++ b/.snippets/webservers/apache.conf
@@ -11,6 +11,7 @@
   php_value upload_max_filesize 100M
   php_value post_max_size 100M
   <Directory "/var/www/pterodactyl/public">
+    Require all granted
     AllowOverride all
   </Directory>
   SSLEngine on


parkervcp:
So the several thousand other installs of the panel. How exactly does this fix what's not broken?

Justman10000:
Like mine?

DaneEveritt:
Please include relevant information as to why the PR should be merged, specifically what it is fixing and why it is necessary. I don't use Apache regularly so a PR like this is just going to be ignored unless you can explain more.

Justman10000:
So ... Apache2 basically locks every directory on the entire operating system, unless you release this lock for directories (or files) with the code given above! In other words, without this code, access to the panel will be denied!

DaneEveritt:
So what specifically about your setup is different though as @parkervcp was asking, since we've had this code for years without people running into issues.

Justman10000:
> 
> 
> So what specifically about your setup is different though as @parkervcp was asking, since we've had this code for years without people running into issues.

So? Strange...

fabm3n:
I can understand @Justman10000 because if you follow the security tips from the Apache docs (https://httpd.apache.org/docs/2.4/misc/security_tips.html#protectserverfiles) you should disallow the access to the whole filesystem and then reallow it again.
So maybe this should be merged (but with the other apache conf also be modified).

parkervcp:
Most people are not following the apache security guidelines. The same goes for any other software out there.

fabm3n:
> Most people are not following the apache security guidelines. The same goes for any other software out there.

Okay yeah that's true and if you are following this guide you should also be able to fix this issue on your own.

PR #370, opened 2021-06-01
mv

Justman10000:
Why copy? Just rename it!


parkervcp:
Because keeping a copy of the default config isn't hurting anybody.

PR #372, opened 2021-06-08
PHP Repo Debian 10 Change

JackW6809:
Personally, this is the only way I can get PHP installed nowadays. Used to work fine the old way before PHP 8 came out but now it just gives me the error php modules are not found and this fixes it for me.

### DIFF
--- a/community/installation-guides/panel/debian10.md
+++ b/community/installation-guides/panel/debian10.md
@@ -25,6 +25,10 @@ systemctl enable mariadb
 
 ### PHP 7.4
 ```bash
+# Add additional repositories for PHP, Redis, and MariaDB
+sudo wget -O /etc/apt/trusted.gpg.d/php.gpg https://packages.sury.org/php/apt.gpg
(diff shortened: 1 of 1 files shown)


i-eat-burgers (dismissed):
Seems fine and worked with me :D
Thanks!

community/installation-guides/panel/debian10.md
```diff
@@ -25,6 +25,10 @@ systemctl enable mariadb
 
 ### PHP 7.4
 ```bash
+# Add additional repositories for PHP, Redis, and MariaDB
+sudo wget -O /etc/apt/trusted.gpg.d/php.gpg https://packages.sury.org/php/apt.gpg
```
Software-Noob, line 29: Sudo is not used in any of the other command examples either. Please remove it. Preferably we'd use curl since it was installed previously

Software-Noob (approved):
I adjusted it to use curl. Thank you for the contribution!

PR #373, opened 2021-06-09
installing.md

GoldTanzil:


JackW6809:
I am not sure what you are trying to achieve here. It actually does not work from my testing so the changes you suggested makes no sense.

```
root@panel:/var/www/pterodactyl# hostname -I | awk '{print $1}'
145.239.115.75
root@panel:/var/www/pterodactyl# tzzone -I | awk '{print $1}'
-bash: tzzone: command not found
```

GoldTanzil:
Leave

PR #379, opened 2021-06-24
docs: move pid limit to the correct section

Software-Noob:


DaneEveritt:
This documentation is super confusing actually.

https://github.com/pterodactyl/wings/blob/develop/config/config_docker.go#L63

This is nested under `docker:` not under the global space?

Software-Noob:
> This documentation is super confusing actually.
> 
> https://github.com/pterodactyl/wings/blob/develop/config/config_docker.go#L63
> 
> This is nested under `docker:` not under the global space?

There are many other values in the same section to show they exist with a link to the files, because users tend to ask if you can change X or Y. 

Do you want to separate them all into different categories?

Software-Noob:
I guess it's fine for now. Will rework it all in V2 to have proper categories.

PR #380, opened 2021-06-25
Update debian9.md

JackW6809:
Fixes an issue where it will automatically install MariaDB 10.1 (Which is not supported, 10.2 is lowest supported version).


Software-Noob (approved):

PR #381, opened 2021-07-06
Update troubleshooting.md

Dungeonseeker:
Updated firewalld command to include adding network interface to docker zone

### DIFF
--- a/panel/1.0/troubleshooting.md
+++ b/panel/1.0/troubleshooting.md
@@ -126,4 +126,11 @@ firewall-cmd --permanent --zone=trusted --change-interface=pterodactyl0
 firewall-cmd --reload
 ```
 
+If you the error `pterdactylnw already assigned to a zone` you have to add the interface to the docker zone instead
+
+```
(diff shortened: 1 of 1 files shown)


panel/1.0/troubleshooting.md
```diff
@@ -126,4 +126,11 @@ firewall-cmd --permanent --zone=trusted --change-interface=pterodactyl0
 firewall-cmd --reload
 ```
 
+If you the error `pterdactylnw already assigned to a zone` you have to add the interface to the docker zone instead
```
Software-Noob, line 129: This contains typos. Are you sure about the error message itself? I am certain it's `ZONE_CONFLICT: 'pterodactyl0' already bound to a zone` for which we have auto-response on our Discord server.

matthewpi:
Adding `docker0` and `pterodactyl0` to the docker zone is no longer required with newer versions of Docker and will prevent Docker from starting if this is added.  While this error may still occur to people on older versions of Docker, it isn't necessary and the solution should be to update Docker.

PR #382, opened 2021-07-10
Update CentOS for AlmaLinux & RockyLinux

markd69:
After testing, this guide works for AlmaLinux and Rocky Linux. The guide should be updated accordingly.


Software-Noob (approved):

Software-Noob (changes requested):
Renaming the file would require updating all the paths that reference it. You can leave the file name be centos8.md.

markd69:
@Software-Noob, I’ve updated the file naming and also changed the wings documentation for community centos 8.

Software-Noob (approved):

PR #383, opened 2021-07-25
Add instructions for downloading Wings for ARM

HandyHat:
Clarifies the need to download a different file


Software-Noob:
It should specify ARM64 since 32-bit is not supported

HandyHat:
> 
> 
> It should specify ARM64 since 32-bit is not supported

Updated :)

Software-Noob (approved):
Thank you for your contribution! 

I made some changes to it, such as separating them into tabs and including them on the wings upgrade page.

PR #384, opened 2021-07-25
Add caddy instructions

HandyHat:
A similar config can be seen here: https://caddy.community/t/nginx-config-translation-help/8166

### DIFF
--- a/.snippets/webservers/caddy-php8.0-nossl.conf
+++ b/.snippets/webservers/caddy-php8.0-nossl.conf
@@ -1,0 +1,20 @@
+http://<domain> {
+    root * /var/www/pterodactyl/public
+
+	header {
+		X-Content-Type-Options nosniff
+		X-XSS-Protection "1; mode=block"
+		X-Robots-Tag none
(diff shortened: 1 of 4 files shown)


Software-Noob:
There should be an option to have both SSL and non-SSL options for users that don't have a domain, since Letsencrypt certificates are only valid for domains. The latter can be achieved by adding `tls off` or specifying `http:// <domain>` for the host.

I have not tried running Laravel app in Caddy itself, so personally, I would have to test first to comment on anything else.

HandyHat:
> 
> 
> There should be an option to have both SSL and non-SSL options for users that don't have a domain, since Letsencrypt certificates are only valid for domains unless ZeroSSL is configured. The latter can be achieved by adding `tls off` or specifying `http:// <domain>` for the host.
> 
> I have not tried running Laravel app in Caddy itself, so personally, I would have to test first to comment on anything else.

I've added instructions for using Caddy without SSL.

WavingCatDevs:
Is this going to be ever merged lol

parkervcp:
I am going to close this in favor of #535

PR #393, opened 2021-08-12
Remove 'FLUSH PRIVILEGES;' calls

eest:
Not necessary when using GRANT:
https://dev.mysql.com/doc/refman/8.0/en/privilege-changes.html


AreYouRlyScared:
Is this still required for MariaDB? If so. It needs to be kept. No harm with it there

eest:
Neither the GRANT-specific example at https://mariadb.com/kb/en/grant/#grant-examples or the more general permission management example at https://mariadb.com/kb/en/mariadb-authorization-and-permissions-for-sql-server-users/#permissions mention any FLUSH usage.

Software-Noob (approved):
This is correct as account statements reload the grants table memory for MariaDB and MySQL for the versions we support. This is most likely leftover from back in the days when INSERT was used to modify the grant tables directly.

Updating the legacy 0.7 guide is unnecessary. It's only there for archive purposes.

PR #394, opened 2021-08-13
Remove unnecessary `USE mysql;` statements

eest:
https://mariadb.com/kb/en/create-user/:
```
For each account, CREATE USER creates a new row in mysql.user (until MariaDB
10.3 this is a table, from MariaDB 10.4 it's a view) or mysql.global_priv_table
(from MariaDB 10.4) that has no privileges.
```

https://dev.mysql.com/doc/refman/8.0/en/create-user.html#create-user-overview:
```
For each account, CREATE USER creates a new row in the mysql.user system table.
```

Follow-up to https://github.com/pterodactyl/documentation/pull/393

### DIFF
--- a/tutorials/mysql_setup.md
+++ b/tutorials/mysql_setup.md
@@ -25,8 +25,6 @@ to our database. You can also use `%` as a wildcard or enter a numeric IP. We wi
 to `somePassword`.
 
 ``` sql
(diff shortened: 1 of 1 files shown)


Software-Noob (approved):

PR #395, opened 2021-08-18
Removed FiveM

AndreasTheDead:
After the latest Problems with FiveM I noticed that it should be removed from the Supportet Game List.

### DIFF
--- a/project/introduction.md
+++ b/project/introduction.md
@@ -27,7 +27,6 @@ and there are plenty more games available provided by the community. Some of the
 * San Andreas: MP
 * Pocketmine MP
 * Squad
-* FiveM
 * Xonotic
 * Discord ATLBot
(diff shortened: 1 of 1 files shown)


parkervcp:
Technically it still works just not out of the box.

AndreasTheDead:
Yes, i know, i thought. iit should be the Point "and More" because you cannot guarantee that it will work in the future.

Software-Noob:
FiveM itself is a community-made egg located in the Parkers repository to which the documentation provides a link. It's not a core official egg of the Panel itself. Personally, I don't see the need for this change. 

> Yes, i know, i thought. iit should be the Point "and More" because you cannot guarantee that it will work in the future.

We can't guarantee that any egg will work in the future. They all require changes and updates at some point.

BenasPaulikas:
#boicot_fivem

PR #396, opened 2021-08-21
Updating to add some notes about SELinux.

ccallahan:
While trying to get Wings to work with Fedora CoreOS 34, I started having a *lot* of AVC denials coming from Docker. After looking into the policy and tagging up with some coworkers, I discovered that especially on Fedora and Fedora-derivatives, including Enterprise Linux, the targeted SELinux policy expects Docker to only read and write data from /var/srv/containers[1] in the context of mounting a directory on the host.

[1] https://github.com/containers/container-selinux/blob/main/container.fc#L137

### DIFF
--- a/community/installation-guides/wings/centos8.md
+++ b/community/installation-guides/wings/centos8.md
@@ -38,3 +38,7 @@ firewall-cmd --reload
 
 ## Installing Wings
 Great, now all of the dependencies and firewall rules have been dealt with. From here follow the [official Wings installation documentation](/wings/1.0/installing.html#enabling-swap).
+
+::: tip
+If you have SELinux enforcement enabled and you are getting AVC denials from your containers, try relocating your Wings data directory from `/var/lib/pterodactyl` to `/var/srv/containers/pterodactyl`. That is where the targeted policy expects Docker to read and write data from.
+:::


matthewpi (approved):

parkervcp:
There is likely to be a flash to add I'm selinux to match the file context

ccallahan:
> There is likely to be a flag to add in selinux to match the file context

I mean, I feel like a better option would be for Wings to check for the presence of /etc/{redhat,fedora}-release and to just make an on the fly configuration change to use /var/srv/containers instead of /var/lib, that way we are more in line with the targeted policy instead of trying to jack with contexts and potentially risk weakening the policy.

parkervcp:
> 
> 
> > There is likely to be a flag to add in selinux to match the file context
> 
> I mean, I feel like a better option would be for Wings to check for the presence of /etc/{redhat,fedora}-release and to just make an on the fly configuration change to use /var/srv/containers instead of /var/lib, that way we are more in line with the targeted policy instead of trying to jack with contexts and potentially risk weakening the policy.

This would fork the config and make it harder to diagnose if someone doesn't know. It would also make our documentation longer winded to explain why it changed for what looks like "no reason".

ccallahan:
> > > There is likely to be a flag to add in selinux to match the file context
> > 
> > 
> > I mean, I feel like a better option would be for Wings to check for the presence of /etc/{redhat,fedora}-release and to just make an on the fly configuration change to use /var/srv/containers instead of /var/lib, that way we are more in line with the targeted policy instead of trying to jack with contexts and potentially risk weakening the policy.
> 
> This would fork the config and make it harder to diagnose if someone doesn't know. It would also make our documentation longer winded to explain why it changed for what looks like "no reason".

So, I respectfully disagree with that. We really aren't forking anything, we're talking about changing three lines in the config that already support being changed. And I don't think it would make the documentation much longer at all. 

But I feel like we're also getting away from the main point, which is that the policy specifies that the kernel only will allow under a strict interpretation of the targeted policy Docker to read, write, and execute data in that particular directory. But in the real world, while I haven't tested Wings in CentOS Stream 9 yet, the only real world case is basically CoreOS, so this whole thing is more of a footnote than anything, but yet a footnote that does need to be mentioned I feel in at least the community docs.

Issue #397, opened 2021-08-22
Update "Enabling swap" part in Wings installtaion instructions


realkarmakun:
Some machines (presumably depending on provider) don't use `GRUB_CMDLINE_LINUX_DEFAULT` and instead only use `GRUB_CMDLINE_LINUX`. This happend to me. I use OVH and even though I have `GRUB_CMDLINE_LINUX_DEFAULT` in grub config, my machine ignores it. 

To figure out what arguments does our machine uses we can type `cat /proc/cmdline` in console and compare it to `/etc/default/grub`. So if we see cmdline arguments from `GRUB_CMDLINE_LINUX` but not from `GRUB_CMDLINE_LINUX_DEFAULT` it's more appropriate to put `swapaccount=1` to `GRUB_CMDLINE_LINUX`. 

I had `WARNING: No swap limit support` on six of my OVH machines (not sure if it's just OVH bare-metal quirk or not). I think this info needs to be included in docs in some way.

PR #398, opened 2021-08-22
Add TIP to swap limit block

realkarmakun:
Resolves https://github.com/pterodactyl/documentation/issues/397

### DIFF
--- a/wings/1.0/installing.md
+++ b/wings/1.0/installing.md
@@ -98,10 +98,11 @@ GRUB_CMDLINE_LINUX_DEFAULT="swapaccount=1"
 ```
 
 ::: warning
-On some systems `GRUB_CMDLINE_LINUX_DEFAULT` can be ignored by the system. So even after manipulations above you will
+On some systems `GRUB_CMDLINE_LINUX_DEFAULT` can be ignored. So even after manipulations above you will
 still see `WARNING: No swap limit support`. So it would be more appropriate to use `GRUB_CMDLINE_LINUX` instead. You
-can figure it out by looking at `cat /proc/cmdline` output and comparing it to `/etc/default/grub`. So if you don't see
-arguments from `GRUB_CMDLINE_LINUX_DEFAULT` in your system you most likely need to use `GRUB_CMDLINE_LINUX`.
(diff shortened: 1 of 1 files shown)


realkarmakun:
I'm not native English speaker, so text probably needs to be modified.

Software-Noob:
`GRUB_CMDLINE_LINUX_DEFAULT` appends entries to the linux line in normal boot mode, while `GRUB_CMDLINE_LINUX` applies to both normal and recovery bot mode.

 I don't see how it would be ignored during normal boot unless this is another one of the OVH quirks or you are in recovery, do you have any source for this?

realkarmakun:
> I don't see how it would be ignored during normal boot unless this is another one of the OVH quirks or you are in recovery, do you have any source for this?

I am kinda my own source, I ran into the issue on OVH bare-metal servers. Though I have been buying servers more and more servers for a year or so and lately there was no quirks around `/home` directory. Even though it used to be that way. But I was updating some of my servers to newer distro version so maybe it's just distribution update issue. I'll try asking in OVH support about this.

realkarmakun:
Sorry for taking so long. Here is supports answer.
Transcript:
me:
```
Additional information:
My /etc/default/grub lacks GRUB_CMDLINE_LINUX_DEFAULT line. 
Is it ignored by OS when using OVH servers? Is it default behavior?
 I'm writing down some documentation, so I'd like to know if this is normal for all OVH bare metal servers.
```
OVH:
```
Refering your request about GRUB configuration files, we confirm that your server seems to be currently installed with Ubuntu20.04 version distribution.

This distribution template has been taken directly from Ubuntu, there has been no modification on the configuration files.
So this means that if your server has a behaviour or specific configuration, it will be the same on all the servers that use this specific distribution.

If you reinstall the server with another distribution, the behaviour or configuration may be different.
```
And some proof:
изображение

Lack of GRUB_CMDLINE_LINUX_DEFAULT is default for Ubuntu 20.04. Can't really check since I'm on windows.

Software-Noob (approved):
Thanks, this tip might save unnecessary trouble for someone. I changed the wording a bit.

In the future, please do not use the `master` branch and instead create a new one. It makes it very annoying to work with.

PR #399, opened 2021-08-23
Add "Hardware requirements" for the panel

Boy132:
Sometimes people on the discord ask what hardware resources are needed to run the panel.

The values given are based on personal experience - if you feel like they are unrealistic let me know.

### DIFF
--- a/panel/1.0/getting_started.md
+++ b/panel/1.0/getting_started.md
@@ -18,6 +18,10 @@ of basic linux system administration you should stop and turn around now.**
 A web server that is just running Pterodactyl will require 2 to 4 GB RAM and 1 to 2 (v)Cores of any half-decent CPU. Approximately 500 MB of disk space will be used.
 These values may vary depending on traffic/ usage and other web applications that might run on the same web server.
 
+::: warning
+This does not include the required resources for wings or any game server!
+:::
+
 ### Picking a Server OS
 Pterodactyl runs on a wide range of operating systems, so pick whichever you are most comfortable using.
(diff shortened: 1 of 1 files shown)


Software-Noob (commented):
Hardware requirements are difficult to provide. Panel can run on the cheapest VPS with 1GB memory or require 8GB+ for larger sites. 

It's a double-edged sword since these requirements can be way overkill for self-hosting a panel. Vcpu for "half-decent" is not very technical when most VPS are just advertised as Vcpu

Boy132:
> Hardware requirements are difficult to provide. Panel can run on the cheapest VPS with 1GB memory or require 8GB+ for larger sites.

Yes I know. This would give at least some rough values to newer users. And thats why I added the little disclaimer that the values might vary.

> It's a double-edged sword since these requirements can be way overkill for self-hosting a panel. Vcpu for "half-decent" is not very technical when most VPS are just advertised as Vcpu

Yeah the panel doesn't require much itself in most cases but I thought this would sound better than just "not much needed".

Software-Noob:
These are still way overkill for self-hosting, which I'd say are the majority of the users. In my opinion, should there be official minimum hardware requirements, they must indeed be the minimum proper requirements and not a guess - the range in your guess is 100% which is huge. Otherwise, there is no real point in displaying it.

I don't think specifying vcpu is even required in this context. All that really matters is the memory and storage. I can confirm that you can run the Panel on the cheapest VPS with 2GB of memory.

PR #404, opened 2021-09-27
acme.sh command is wrong and doesn't work with DNS

thierrybla:
The row "acme.sh --issue --standalone -d "example.com" --dns dns_cf \" should be changed to "acme.sh --issue --dns dns_cf -d "example.com" \" otherwise it wont use DNS and it will require port 80 to be open.


Software-Noob (approved):

PR #405, opened 2021-10-06
Add stopping wings to update manual

Jelcoo:
This is to prefent `Warning: Failed to create the file /usr/local/bin/wings: Text file busy` when updating.

### DIFF
--- a/wings/1.0/upgrading.md
+++ b/wings/1.0/upgrading.md
@@ -16,8 +16,15 @@ most cases your base Wings version should match that of your Panel.
 | 1.5.x         | 1.4.x         |           |
 | **1.6.x**     | **1.5.x**     | ✅        |
 
+## Stopping Process
+First, stop the running wings instance.
+
+```bash
+systemctl stop wings
+```
+
 ## Download Updated Binary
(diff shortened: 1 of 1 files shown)


TASelwyn:
What OS are you running?

I've updated wings dozens of time with just running the download command, chmod+x'ing it then restarting the wings service.
I've never once encountered an issue with the wings file being locked/busy.

Jelcoo:
> What OS are you running?
> 
> I've updated wings dozens of time with just running the download command, chmod+x'ing it then restarting the wings service. I've never once encountered an issue with the wings file being locked/busy.

This didn't happen to me, but to someone I know. It doesn't have to happen all the time, but it is probably better to stop wings, so the panel doesn't send requests to wings and it handles it half because you just updated it.

matthewpi:
I do see this error in my dev environment when I use `scp` to upload a new wings binary while the wings service is already running.

Software-Noob:
@Jelcoo Can you please resolve the conflicts? To keep the update guide short, you can add the stop command above the curl one.

It appears a lot of users are experiencing this issue on Discord.

PR #416, opened 2021-11-30
Explicitly tell to delete default webserver config

Boy132:
Many users forget to delete the default config and then ask on the Discord for help. Seems like the big red "DANGER" at the top isn't explicit enough.

### DIFF
--- a/panel/1.0/webserver_configuration.md
+++ b/panel/1.0/webserver_configuration.md
@@ -1,23 +1,17 @@
 # Webserver Configuration
 
-## Removing the default configuration
-First of all you should remove the default Apache or NGINX configuration as it will expose application secrets to malicious users by default.
-``` bash
-# If using NGINX:
-rm /etc/nginx/sites-available/default && rm /etc/nginx/sites-enabled/default
-
-# If using Apache:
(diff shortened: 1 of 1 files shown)


Software-Noob:
You only have to remove the symlinked from sites-enabled for Nginx to disable it. Disabling it on Apache is instead done using `a2dissite 000-default.conf` that'll work no matter the install path. The web server must be restarted to apply this, which many will probably miss.

These instructions would better fit in the respective tabs below for Apache and Nginx instead of showing globally for both. Default site showing up is also a configuration issue, meaning the server name does not match, and thus it redirects to the default one

Boy132:
> You only have to remove the symlinked from sites-enabled for Nginx to disable it. Disabling it on Apache is instead done using `a2dissite 000-default.conf` that'll work no matter the install path. The web server must be restarted to apply this, which many will probably miss.
> 
> These instructions would better fit in the respective tabs below for Apache and Nginx instead of showing globally for both. Default site showing up is also a configuration issue, meaning the server name does not match, and thus it redirects to the default one

I made some changes based on your feedback! :)

Software-Noob (approved):

PR #417, opened 2021-12-05
add the exit command to the panel mysql setup.

QuintenQVD0:
In blue box for setting up the mysql user for the panel add the exit command so you go back to you normal shell.

afbeelding


Software-Noob (approved):

PR #418, opened 2021-12-10
docs(wings): Display ARM64 download for wings

Boy132:
Often people with arm64 don't realize that they have to download the arm variant of wings because it is nowhere noted.


Software-Noob:
We had this a while ago in separate tabs, but there was a hydration issue causing the whole modal to disappear occasionally, and I completely forgot to look further into it.

The way you are suggesting to do it is most likely going to cause someone on AMD64 to install the ARM64 version by copying all the commands. 

image

I think they should be separated to prevent such mistakes because people don't tend to read that carefully.

image


Ideally, we'd have them under tabs, but that's probably something to do in vuepress V2 rewrite.

Boy132:
> I think they should be separated to prevent such mistakes because people don't tend to read that carefully.

Ah yes, users not reading carefully. 🙄 

Anyways, I made the changes you suggested.

Software-Noob (approved):

PR #420, opened 2021-12-13
docs(wings): Show tip to figure out architecture

Boy132:
Follow up to #418.
Seems like the other PR confuses some people on what they need to use. This tip should help.

### DIFF
--- a/wings/1.0/installing.md
+++ b/wings/1.0/installing.md
@@ -103,7 +103,7 @@ Some Linux distros may ignore `GRUB_CMDLINE_LINUX_DEFAULT`. Therefore you might 
 The first step for installing Wings is to make sure we have the required directory structure setup. To do so,
 run the commands below which will create the base directory and download the wings executable.
 
-::: tip Run `uname -m` to figure out what build you need. :::
+::: tip Run `uname -m` to figure out what build you need. Unless it returns something like `arm` or `aarch` you need to use the AMD64 build. :::
 
 ### For AMD64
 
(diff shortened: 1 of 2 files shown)


DaneEveritt:
@Software-Noob is there a reason we can't use tabs here?

Boy132:
> @Software-Noob is there a reason we can't use tabs here?

Aren't tabs for these subtle differences between the two instructions a bit "overkill"? Sure they would look nicer but the content would be 99% the same, wouldn't it?

Software-Noob:
> @Software-Noob is there a reason we can't use tabs here?

We could and should, but last PR of tabs caused hydration issue occasionally that made the whole tab modal disappear and thus break whole documentation.

I could not figure out why nor replicate it in my own setups. Can't really test on master docs since it's a protected branch and reviews of new PRs takes time

DaneEveritt:
I think this current format is unnecessarily confusing, only a small percentage of users need the ARM build. Perhaps just drop the whole section for downloading it, and just make a tip box that lets a user know there is an arm specific build and to replace `wings_linux_amd64` with `wings_linux_arm64`.

Software-Noob:
I agree that they're a bit confusing right now.  We could use something like this instead?

```bash
mkdir -p /etc/pterodactyl
curl -L -o /usr/local/bin/wings https://github.com/pterodactyl/wings/releases/latest/download/wings_linux_$([[ "$(uname -m)" == "aarch64" ]] && echo "arm64" || echo "amd64")
chmod u+x /usr/local/bin/wings
```

PR #422, opened 2021-12-20
Update getting_started.md

Jcodeerd:
remove the panel.tar.gz file, because there is no use in keeping it


parkervcp:
Keeping the tar around could be helpful for people twerking the panel and wanting to revert back to stock.

Jcodeerd:
Yep, thats fair.

PR #424, opened 2021-12-23
Permissions and removal of the archive

Justman10000:
Better to use 755 on the entire directory, in addition, it would be unnecessary to leave the .tar.gz archive lying around!

### DIFF
--- a/panel/1.0/getting_started.md
+++ b/panel/1.0/getting_started.md
@@ -107,9 +107,9 @@ please have a look at [Setting up MySQL](/tutorials/mysql_setup.html).
 mysql -u root -p
 
 # Remember to change 'yourPassword' below to be a unique password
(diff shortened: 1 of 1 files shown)


Software-Noob:
#422 regarding panel tar. Should someone want to remove it, they surely can do it themselves.

There's no reason to set 755 on all files. Files should have the minimum required permissions in production, especially files containing secrets. Also, blindly setting it to `.` instead of an absolute path can have dire consequences if the user is not in the expected directory - this happens often.

localhost and 127.0.0.1 are not the same. The former involves name resolution and can have a [different effect regarding](https://dev.mysql.com/doc/refman/8.0/en/connecting.html) connection using sockets vs. TCP. Also, you made a typo in it.

PR #425, opened 2021-12-23
Only from PHP 8.0

Justman10000:
I had Debian 11 and PHP 7.4 on it, only after uninstalling PHP 7.4 and installing PHP 8.0 did it work!


Software-Noob:
7.4 and 8.0 are both supported. Distro version doesn't affect the PHP requirements

Justman10000:
> 

Well, for me it only works with PHP 8.0...

hwalker928:
Are you sure it wasn't just an NGINX configuration issue?

Software-Noob:
That would not be up to the distro and rather your configuration. Most likely your web server virtual host was set to php 8.

Both php 7.4 and 8.0 are supported as listed in the [required dependencies](https://pterodactyl.io/panel/1.0/getting_started.html#dependencies)

PR #426, opened 2021-12-29
Fix recaptcha docs typo

VersatileSft:
"If users have trouble logging in, or your Panel isn't exposed to the internet, in can make sense to disable reCAPTCHA."
This sentence is grammatically incorrect, it looks like "it" was misspelt as "in".


Software-Noob (approved):

PR #431, opened 2022-01-27
Add sudo where necessary

Term14:
Added `sudo` to the beginning of any command that requires it to function when used with a user account with `sudo` privileges.

Without including `sudo` new users may be tempted to, or believe it's necessary, to switch to the root user. This definitely isn't something that should be done so best that the commands that need privilege escalation say so.


Software-Noob:
Thank you for your contribution. Unfortunately, you've removed the origin branch and as such, I can't leave any reviews or modify it to apply proposed changes. Please feel free to restore the branch from the commit or open a new PR keeping your fork alive until it has been merged.

PR #437, opened 2022-02-15
Ubuntu 20.04: command to add PPA for PHP 8

pkirkovsky:
This PPA is required to install PHP 8 on Ubuntu 20.04


Software-Noob:
Thank you, but this guide is hidden and not used. The official documentation is written for Ubuntu and works for all LTS releases.

PR #443, opened 2022-04-17
Add updating composer warning to prevent wanting death when you realise it broke everything 2 hours later

BumbleTree:
I found this out the hard way... And yet there is no disclaimer for it.


DaneEveritt:
Seeing as we never tell you to run it...

PR #445, opened 2022-04-22

GoedendagMC:
Updated syntax, --key-file is deprecated.

### DIFF
--- a/tutorials/creating_ssl_certificates.md
+++ b/tutorials/creating_ssl_certificates.md
@@ -94,8 +94,8 @@ Then create the certificate.
 
 ```bash
 acme.sh --issue --dns dns_cf -d "example.com" \
---key-file /etc/letsencrypt/live/example.com/privkey.pem \
---fullchain-file /etc/letsencrypt/live/example.com/fullchain.pem 
+--keypath /etc/letsencrypt/live/example.com/privkey.pem \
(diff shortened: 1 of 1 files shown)


Software-Noob:
This is actually the opposite. Your proposed parameters were[ deprecated in 2017](https://github.com/acmesh-official/acme.sh/commit/5c539af7d7645723b928389bc93da526f0dcfa60) and there's backward compatibility left in acme.sh to support both as seen [here ](https://github.com/acmesh-official/acme.sh/blob/master/acme.sh#L7384-L7399)and [here](https://github.com/acmesh-official/acme.sh/blob/master/acme.sh#L5684). The official examples also use -`-key-file` and `--fullchain-file`

Could you please provide a source for where you got this information from, as I can't find it?

GoedendagMC:
The simple reason is that the other arguments didn't work for me, yet these did. I had the latest acme.sh installed though.

Software-Noob:
I've tested and they're working on Acme V2 and V3. The existing commands are the same as official [acme examples on Github](https://github.com/acmesh-official/acme.sh#3-install-the-cert-to-apachenginx-etc) and listed in the [help parameters](https://github.com/acmesh-official/acme.sh/blob/master/acme.sh#L6800-L6805) `acme.sh --help`. For that reason, I'm closing this request. 

Thank you for trying to contribute 👍

PR #447, opened 2022-05-02
docs: instruct to enable ssl module for apache

Boy132:
Running `a2enmod ssl` will prevent the following error that often people on the discord have:
`Invalid command 'SSLEngine', perhaps misspelled or defined by a module not included in the server configuration`


Software-Noob (approved):
LGTM. Thank you for contributing :+1:

PR #448, opened 2022-05-08
add ubuntu 22.04

QuintenQVD0:
This is my first time ever using vuepress or writing any sort of documentation so please let me now if anything has to be changed or this is not relevant.


Software-Noob (commented):
There is no need for a separate community guide. I have it added in a pending branch for Pterodactyl 1.8

PR #449, opened 2022-05-08
Update documentation for 1.8 release

Software-Noob:
#### Added

Add Ubuntu 22.04 to the Guide

#### Changed
Use official Redis PPA, which supports all of the supported Ubuntu/Debian versions.
Use PHP 8.1, which for newer distro sources will be the default on apt upgrade and currently causes headaches for some users
Use Pterodactyl 1.8 for the Panel version.

Tested on Ubuntu LTS releases 18, 20, and 22.

### DIFF
--- a/guides/php_upgrade.md
+++ b/guides/php_upgrade.md
@@ -41,7 +41,7 @@ is most likely called `pterodactyl.conf` and located in the `/etc/nginx/sites-av
 Make sure to update the path in the command below to reflect the actual location of your configuration file.
 
 ``` bash
-sed -i -e 's/php7.[0-4]-fpm.sock/php8.1-fpm.sock/' /etc/nginx/sites-available/pterodactyl.conf
+sed -i -e 's/php[7|8].[0-9]-fpm.sock/php8.1-fpm.sock/' /etc/nginx/sites-available/pterodactyl.conf
 ```
 
 Once you have edited the file run the command below to reload nginx and apply your changes.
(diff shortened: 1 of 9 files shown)


guides/php_upgrade.md
```diff
@@ -33,33 +34,36 @@ composer self-update --2
...
-After upgrading to PHP 8.0, you will most likely need to update your NGINX configuration. Your configuration file
+After upgrading to PHP 8.1, you will most likely need to update your NGINX configuration. Your configuration file
 is most likely called `pterodactyl.conf` and located in the `/etc/nginx/sites-available/` directory, or if on CentOS,
 `/etc/nginx/conf.d/`.
 
 Make sure to update the path in the command below to reflect the actual location of your configuration file.
+
 ``` bash
-sed -i -e 's/php7.[0-4]-fpm.sock/php8.0-fpm.sock/' /etc/nginx/sites-available/pterodactyl.conf
+sed -i -e 's/php7.[0-4]-fpm.sock/php8.1-fpm.sock/' /etc/nginx/sites-available/pterodactyl.conf
```
DaneEveritt, line 44: Shouldn't this look for `php[7|8].[\d]-fpm.sock`? Seems like we could match even existing 8.1 installs at this point, not a big deal.

guides/php_upgrade.md
```diff
@@ -33,33 +34,36 @@ composer self-update --2
...
+Run the commands below to disable PHP 7.4 and enable PHP 8.1 when serving requests. If you are upgrading from
 PHP 7.4 change the value in the command below to reflect that.
 
 ``` bash
 # Hint: a2enmod = a2_enable_module 🤯
-a2enmod php8.0
+a2enmod php8.1
 
 # Hint: a2dismod = a2_disable_module 🤯
 a2dismod php7.3
```
DaneEveritt, line 63: Disable 7.3 upgrading from 7.4? We should cover 8 as well?

DaneEveritt (changes requested):
Not sure if we're bumping Wings to 1.6, need to review the changes, I don't think there is anything significantly different that would break in Panel@1.7.

guides/php_upgrade.md
Software-Noob: Changed to a wildcard, so we'll disable any PHP versions users might have without guessing what it might be before enabling 8.1

guides/php_upgrade.md
Software-Noob: You're right, I missed this.  `\d` doesn't exist in sed for regular expression as it's a switch. We can use `0-9` or `\w.`. The first is probably more foolproof as it matches digits only?

Software-Noob:
> Not sure if we're bumping Wings to 1.6, need to review the changes, I don't think there is anything significantly different that would break in Panel@1.7.

Wings are already on 1.6 and the new Panel release would be 1.8?

PR #450, opened 2022-05-19
Add "composer update" warning

Boy132:
This is based on #443.

Dane said:
> Seeing as we never tell you to run it...

Actually when running `composer install` for the first time it suggests to update the dependencies. There is no harm in showing a little warning in the docs.

### DIFF
--- a/panel/1.0/getting_started.md
+++ b/panel/1.0/getting_started.md
@@ -132,6 +132,10 @@ composer install --no-dev --optimize-autoloader
 php artisan key:generate --force
 ```
 
+::: warning
+Do not run `composer update` or you will break your panel!
+:::
+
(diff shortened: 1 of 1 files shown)


cmrxnn:
Adding more detail to the message in order to explain the issue might be beneficial to the end user as they'll understand _why_ they shouldn't.

Something like: Due to the Panel designed to use specific dependency versions for Composer, we strongly suggest you do not update these dependencies.

DaneEveritt:
People don't read the documentation thoroughly as is, I'm not interested in adding more warnings and visual distractions just to tell people to not do something we never even mention.

It is annoying, but not worth making a mess of the documentation over.

PR #452, opened 2022-06-03
Also remove the self-upgrade from the artisan CLI documentation for the time being

TsjipTsjip:


Software-Noob:
This documents the existing artisan commands and is not used as a guide for updating. As such, this notice in the current format is unnecessary and also, links to the wrong place (this is located in /community)

PR #453, opened 2022-06-07
use php8.1-common

Tealk:
Note, selecting 'php8.1-common' instead of 'php8.1-pdo'
Note, selecting 'php8.1-common' instead of 'php8.1-tokenizer'

php8.1-common is a meta-package that installs most of the widely used PHP extensions in one-go. It automatically installs package such as php8.1-pdo, php8.1-tokenizer among other useful extensions.

### DIFF
--- a/panel/1.0/getting_started.md
+++ b/panel/1.0/getting_started.md
@@ -69,7 +69,7 @@ apt update
 apt-add-repository universe
 
 # Install Dependencies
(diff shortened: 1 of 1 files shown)


Software-Noob:
It contains more than the two extensions from our list of requirements than what you've replaced. Personally, I prefer to install exact packages or extensions to know what is required is indeed installed. Should the package be in php-common, it will skip it. No harm done in the current method.

Leeo97one:
Except that `php8.1-common` is not a meta-package, it actually is a real package which contains the mentioned extensions. The `php8.1-pdo` and `php8.1-tokenizer` packages no longer exist (they are automatically replaced by `php8.1-common`).

DaneEveritt (approved):

PR #462, opened 2022-06-30
Update upgrading.md

LilSadPanda:
Updated upgrading.md under `Download Updated Binary`


Software-Noob:
Thank you for the contribution. Unfortunately, this change is unneccesary.

The binary is directly downloaded into the directory if you read the curl command.

PR #464, opened 2022-07-02
Create dev-env-windows.md

JensvandeWiel:
A simple guide to customize the panel on windows 10/11


Software-Noob:
Thank you for the contribution, however, Windows is not a supported platform and these are the same instructions present in the installation documentation.

PR #466, opened 2022-07-09
Update installing.md

oktkas:
Add crucial proxy information


Software-Noob:
Please target the [Wings additional configuration](https://github.com/pterodactyl/documentation/blob/master/wings/1.0/configuration.md) page instead for an example vhost for proxying Wings and websocket, as this information is not relevant to most users and will clutter or confuse them during the installation process. 

Also, you shouldn't modify Wings config directly. Once Wings boots it will sync with the Panel and overwrite those port and SSL-enabled settings. They must be updated from the Node settings on the panel itself.

PR #467, opened 2022-07-10
Restart Wings in troubleshooting steps

BakiDance:
Add a line explaining the possible need to restart wings following a certificate renewal. This is already mentioned by the Support Bot in the Pterodactyl Discord server when linking this page, but is neglected on the page itself. Adding this line would help those who use this site as a primary resource. When I followed the steps on this page, I ran into instability with Wings as I didn't realize the certificate didn't apply automatically; having that explained here would have helped.


Software-Noob (approved):

PR #469, opened 2022-07-22
added Install Dependencies if using apache2 as webserver

JensvandeWiel:


parkervcp:
@JensvandeWiel If you are going to add that you should be adding all the other documentation around apache as well. Such as configuration and management like we already do for nginx. 

There is a reason we limited it to a single software package.

Software-Noob:
Copy pasting the same thing with one package difference clutters the documentation. Should someone wish to use another web server, I hope they know how to install it.

Software-Noob:
Thank you for attempting to contribute, but I'm closing this due to the points mentioned above. It'll clutter the documentation too much for a single package.

PR #477, opened 2022-08-17
Add proxy instructions

TheRealGramdalf:
For pr with pterodactyl/documentation

### DIFF
--- a/guides/proxy.md
+++ b/guides/proxy.md
@@ -1,0 +1,30 @@
+# Installing Wings behind a proxy
+
+If you are installing wings behind a reverse proxy, here are some additional tips to help you get started.
+
+## To SSL or not to SSL?
+
+If *at any point* you use an HTTPS connection, pterodactyl will think you are using SSL. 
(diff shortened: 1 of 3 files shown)


Software-Noob:
We already have reverse proxy documentation for the panel at https://pterodactyl.io/panel/1.0/additional_configuration.html#reverse-proxy-setup and having two different type of guides would be confusing. 

Should you wish to include additional configuration documentation for wings, it's something that can be added under wings additional configuration category and not as a seperate guide. That way everything is found in one centralized location from the sidebar.

Altough this guide doesn't appear to contain anything pterodactyl specific. It's fairly barebone reverse proxy information.

TheRealGramdalf:
I'll reformat the guide under wings additional config. The main reason I would do so is that yes, it is a barebones guide, but for someone like me who is fairly experienced in other self-hosted services, but not necessarily aware of all the ins and outs of reverse proxying it is actually extremely helpful. 

It took me a long time to figure this out, since a lot of the issues related to it were vague, which I assume is mainly due to my inexperience. In any case, I'll reformat it.

PR #478, opened 2022-08-17
fix: markdown container formatting

Goksi:


Software-Noob (approved):
Thank you for contributing :+1: 

---

There were some issues that I fixed. For future reference: 

Opening and closing tags `:::` must be on separate lines. The opening tag specifies the container type and additional text on the same line will be the title. Text content must be on a separate line from the tags.

Your attempt set the rest of the file to be inside the warning container. 

[See vuepress docs for more.](https://v1.vuepress.vuejs.org/guide/markdown.html#custom-containers)

PR #480, opened 2022-08-19
Update adding the docker user

playaz44:
Based on https://stackoverflow.com/questions/27701930/how-to-add-users-to-docker-container 
Original ```adduser -D -h /home/container container``` returns error ```Option d is ambiguous (debug, disabled-login, disabled-password)```.

### DIFF
--- a/community/config/eggs/creating_a_custom_image.md
+++ b/community/config/eggs/creating_a_custom_image.md
@@ -57,7 +57,7 @@ single [`RUN`](https://docs.docker.com/engine/reference/builder/#run) block.
 Within this `RUN` block, you'll notice the `useradd` command.
 
 ```bash
-adduser -D -h /home/container container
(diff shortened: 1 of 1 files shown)


parkervcp:
This varies by distribution and `adduser` release version.

playaz44:
I checked on ```arm64v8 debian:bullseye-slim``` maybe it would be worth adding such information.

Software-Noob:
That's why verbose parameters should be preferred, such as `--disable--pasword`; otherwise, the definition can vary, e.g for your case it'd be `-d` and not `-D`. Passing `-s` or `--shell` in this case does nothing as it's empty and is also determined by the entrypoint.

I've always followed the use of `-m -d` with useradd, which creates and changes the home directory. Password is disabled by default when not passing it using useradd.

`useradd --home --create-home /home/container container`

PR #481, opened 2022-08-28
add guide for debian 11 + EOL debian 9 + mention ubuntu 22.04 as supported

QuintenQVD0:
- remove support debian 9 (EOL)
- add ubuntu 22.04 to the supported list
- write a special panel install guide for debian 11


QuintenQVD0:
I have made a new VM with those docs and it works fine

community/installation-guides/panel/debian11.md
```diff
@@ -0,0 +1,65 @@
...
+```bash
+apt -y install software-properties-common curl ca-certificates gnupg lsb-release sudo wget
+```
+
+### php 8.1
+```bash
+# Add repository for PHP
+echo "deb https://packages.sury.org/php/ $(lsb_release -sc) main" | sudo tee /etc/apt/sources.list.d/sury-php.list
+
+wget -qO - https://packages.sury.org/php/apt.gpg | sudo apt-key add -
```
Software-Noob, line 20: Adding `wget` package into the mix is unnecessary. Curl is used everywhere and most of the time preinstalled, so please use it. gnupg2 is generally used as well instead of the classic one.

`apt-key `is also deprecated. The signing key can be added and converted from PGP into GPG with `sudo gpg --dearmor -o <keyring file path>`

Should I remember Debian paths right it would be:

`curl -fsSL  https://packages.sury.org/php/apt.gpg| sudo gpg --dearmor -o /etc/apt/trusted.gpg.d/sury-keyring.gpg`

Software-Noob (changes requested):

Software-Noob (approved):

I've made some necessary style and config changes.

PR #490, opened 2022-10-11
Change TimeZone in .env

Zestyclose-Thanks-82:
Hi,

I created a Minecraft server on Oracle Cloud on Ubuntu with a Pterodactyl panel but on the panel the time is 2 hours behind, so I want to change the timezone in .env but I don't know how I get in the .env file from Ubuntu with PuTTY.


Software-Noob:
Please use the Discord server or Google for Linux support. Basic Linux knowledge is a requirement to manage the panel.

PR #491, opened 2022-10-15
feat(customization): include openssl-legacy-provider notice

### LINKED ISSUE #484: Document customizing with Node v15+
When following the [building panel assets](https://pterodactyl.io/community/customization/panel.html) guide in the documentation, users are likely to encounter a crypto/hashing error if they use NodeJS v15 or above. The documentation should specify setting `NODE_OPTIONS` to `--openssl-legacy-provider` to prevent this (and in [BUILDING.md](https://github.com/pterodactyl/panel/blob/develop/BUILDING.md)).

devnote-dev:
Adds notice for building with NodeJS v15+, closes #484.

### DIFF
--- a/community/customization/panel.md
+++ b/community/customization/panel.md
@@ -37,10 +37,11 @@ yarn # Installs panel build dependencies
 
 ## Build Panel Assets
 
-The following command will rebuild the Panel frontend.
+The following command will rebuild the Panel frontend. For NodeJS version 15 and above, you must enable the `--openssl-legacy-provider` option before building.
 
 ```bash
 cd /var/www/pterodactyl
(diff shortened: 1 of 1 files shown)


Software-Noob:
This should only affect Node +17 (non-lts) which uses OpenSSL 3.0 as default? The v16 was supposed to have it but didn't fit the release schedule. v15 is EOL and had OpenSSL 1.1.1 same as v16.

Ref https://nodejs.org/es/blog/release/v17.0.0/#openssl-3-0

devnote-dev:
@Software-Noob Yeah that's right, I was following incorrect information from the server, my bad. Fixed in latest commit.

PR #496, opened 2022-11-06
docs(s3): document storage class configuration

MatweyCat:
Added description of AWS_BACKUPS_STORAGE_CLASS option. It was added in version 1.10.3 but has not been described.


Software-Noob (approved):

PR #497, opened 2022-11-16
fixed text with 'composer v2'

kartikay-agarwal:


Software-Noob:
Thanks for taking the time to contribute to the project. The dependency is `composer` with the specific version being specified afterward. 

I think enclosing both of them in a code tag looks worse. For this reason, unfortunately I'll be closing the PR.

PR #501, opened 2022-12-07
Proxying Wings FQDN through Cloudflare

DemsarMa:
Added some more info for proxying Wings FQDN through Cloudflare. This has been tested a few minutes ago and came to a conclusion that the FQDN must have the port that Cloudflare has (so HTTPS ports with caching enabled).
Tell me if anything is wrong here, I'm glad to help figure it out!

### DIFF
--- a/wings/1.0/configuration.md
+++ b/wings/1.0/configuration.md
@@ -55,7 +55,7 @@ After making changes, the following commands will stop the Wings, remove the Pte
 
 Cloudflare proxying of the Wings isn't beneficial since users will be connecting to the machine directly and bypassing any Cloudflare protection. As such, your Node machine IP will still be exposed.
 
-To enable Cloudflare proxy, you must change the Wings port to one of the Cloudflare HTTPS ports, such as 8443, because Cloudflare only supports HTTP on port 8080. Select your Node in the Admin Panel, and on the settings tab, change the port.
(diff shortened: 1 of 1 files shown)


Software-Noob:
Thanks for taking the time to contribute to the project 👍 

However, I don't see the requirement for caching? That's for the user to decide unless I am missing something, in which case please correct me. The example port 8443 does support both options.

DemsarMa:
> don't see the requirement for caching?

I've tried other ports populated in ports without caching, resulting in panel not able to detect Wings, if it's behind Cloudflare (orange cloud when assigning the DNS setting). Another reason to think why it'd be best to specify why caching port needs to be used: DNS search services can obviously track those records, including the Wings FQDN. I value the security of other users, so that's why it's best to use that method (Caching port specifically specified).


> The example port 8443 does support both options

Yes correct, however if you want to host Ptero from home (which most does), some ISP's doesn't allow 8443 to be opened (I have no honest idea why would they do that), that happened in my case, that's why I had to use other port that's similar to 8443. I included that link in the PR, so people won't be looking for it, probably asking why they can't use 8443 and have to use a different port (rarely but I'm sure it'll happen to anyone else too).

I could be wrong again what I said, but I did my best to address the situation.

Software-Noob (approved):

PR #502, opened 2022-12-07
Podman section

b4dsk:
Matthew was going to add this so I might as well do it before him


Software-Noob (changes requested):
Thanks for taking the time to contribute to the project :thumbsup: 

The only relevant section for Podman here is the log driver. Podman supports `json-file` and `journald` as far I recall. Be aware that Podman support is also still very much experimental.


I'd keep it simple.

```
## Podman Compatability

Wings v1.10+ has experimental Podman compatibility. The logging driver must be changed to one of the supported Podman options.

docker:
  ...
  log_config: json-file
  ...
 ```

 
The beginning of the documentation specifies that all key values are to be changed in the config file. Repeating it again is not necessary.

PR #503, opened 2022-12-08
Update tab name of acme method in ssl docs

Boy132:
Some users think that it is required to use acme when using cloudflare proxy and generating a ssl cert because of the missleading tab name.
The new name should make it clearer.


matthewpi (approved):

Software-Noob (approved):

PR #509, opened 2022-12-28
update docs for minecraft proxy

parkervcp:
update the IP to allocate from 127.0.0.1 to 172.18.0.1 for the game servers


matthewpi:
This should not be changed https://github.com/pterodactyl/wings/blob/develop/environment/docker/container.go#L158

PR #513, opened 2023-01-05
Updating Documentation to Include % Reference Acct

JosTheDude:
Updating Documentation to also include % as a reference ip for granting privileges on a new database node


Software-Noob:
This defeats the purpose of the first user creation command.

You can reference % as being an option - or even better only restrict the user to panel machine IP.

JosTheDude:
> This defeats the purpose of the first user creation command.
> 
> You can reference % as being an option - or even better only restrict the user to panel machine IP.

Main issue was the fact that I straight up couldn't add a database node without running those extra commands on top of the included ones to create a permissions mysql user, however you are correct in that so c:

PR #518, opened 2023-01-26
add lsb-release to apt install for redis

QuintenQVD0:
We now install redis server with:
```bash
curl -fsSL https://packages.redis.io/gpg | sudo gpg --dearmor -o /usr/share/keyrings/redis-archive-keyring.gpg
echo "deb [signed-by=/usr/share/keyrings/redis-archive-keyring.gpg] https://packages.redis.io/deb $(lsb_release -cs) main" | sudo tee /etc/apt/sources.list.d/redis.list
```

But for this to work lsb-release is needed as it is not installed by default on ubuntu and will not replace: `$(lsb_release -cs)`

### DIFF
--- a/panel/1.0/getting_started.md
+++ b/panel/1.0/getting_started.md
@@ -53,7 +53,7 @@ operating system's package manager to determine the correct packages to install.
 
 ``` bash
 # Add "add-apt-repository" command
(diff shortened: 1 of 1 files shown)


parkervcp (approved):

QuintenQVD0:
it seems to just work for most users to this is not needed anymore

PR #525, opened 2023-03-15
Remove colons from Permissions

PadowYT2:
1. There's no colon for Apache on CentOS
2. Is there really a reason for them?


Software-Noob:
There's no specific use. It's a style choice, but since it doesn't appear to be used anywhere else I guess it's fine to remove.

Software-Noob (approved):

PR #530, opened 2023-05-15
Update installing.md

Loki-101:
Added a small section about DNS since I do see that brought up in the Discord Server often enough to warrant a quick mention here.

### DIFF
--- a/wings/1.0/installing.md
+++ b/wings/1.0/installing.md
@@ -140,6 +140,16 @@ sudo wings --debug
 
 You may optionally add the `--debug` flag to run Wings in debug mode.
 
+### Configuring DNS (Usually not needed, refer back to this if your containers have no internet connection)
+Now that Wings has run successfully and you have gotten the green heart on your Nodes page, the wings config at '/etc/pterodactyl/config.yml' will have new values.
+One of those values is DNS, which by default will be 1.1.1.1 and 1.0.0.1
+For your containers to have internet, your DNS Servers in your wings config MUST match your DNS Servers on your host system.
+To view what DNS Servers your host system is using, you can use systemd-resolve, which typically will be on any system that uses systemd.
+```bash
+systemd-resolve --status | grep "DNS Servers"
+```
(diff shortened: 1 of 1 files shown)


wings/1.0/installing.md
```diff
@@ -140,6 +140,16 @@ sudo wings --debug
 
 You may optionally add the `--debug` flag to run Wings in debug mode.
 
+### Configuring DNS (Usually not needed, refer back to this if your containers have no internet connection)
+Now that Wings has run successfully and you have gotten the green heart on your Nodes page, the wings config at '/etc/pterodactyl/config.yml' will have new values.
+One of those values is DNS, which by default will be 1.1.1.1 and 1.0.0.1
+For your containers to have internet, your DNS Servers in your wings config MUST match your DNS Servers on your host system.
```
parkervcp, line 146: The DNS in the container does not need to match the hosts DNS configuration. The place where this is an issue is for hosts that block the cloudflare ip addresses.

This is most likely due to old configurations or some thought that it makes things more secure.

wings/1.0/installing.md
```diff
@@ -140,6 +140,16 @@ sudo wings --debug
...
+To view what DNS Servers your host system is using, you can use systemd-resolve, which typically will be on any system that uses systemd.
```
parkervcp, line 147: You should specify that `systemd-resolve --status` is the command to run if they want to see their DNS servers.

It's also important to know that not all distributions use `systemd-resolve` to manage their dns.

parkervcp (changes requested):

Loki-101:
Thank you. I think it's clearer now, and I've removed the included filters so just lists all the information, as I don't know how well the filters would work with multiple network interfaces in use.

Boy132:
IMO this should be moved to the "Troubleshooting" section or to "Additional Configuration" because as parker mentioned not everyone has this problem.

Loki-101:
Good point. I'll close this and make a request after moving this section to troubleshooting tomorrow.

PR #537, opened 2023-07-17
Add paragraph differentiating a MySQL database vs a MySQL database instance.

natereprogle:
The term "database" is ambiguous when it comes to MySQL, because many people call an entire _instance_ a database, while instead MySQL databases are what the MySQL database management engine handles. This proposed change notes that ambiguity to hopefully reduce confusion

### DIFF
--- a/tutorials/mysql_setup.md
+++ b/tutorials/mysql_setup.md
@@ -7,6 +7,10 @@ MySQL is a core component of Pterodactyl Panel but it can be confusing to setup 
 This is a very basic tutorial that skims just enough of the surface to set MySQL up and running with the panel.
 If you're interested in learning more, there are some great tutorials available on the Internet.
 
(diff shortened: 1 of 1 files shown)


parkervcp:
@natereprogle 

I would not say the term database is ambiguous. MySQL/MariaDB are database servers/hosts that serve databases to clients.

natereprogle:
> @natereprogle 
> 
> 
> 
> I would not say the term database is ambiguous. MySQL/MariaDB are database servers/hosts that serve databases to clients.

In a recent discussion I had with someone, it seemed there was confusion about Pterodactyl being able to create net-new MySQL database servers, and this article was listed as a source. I figured an aside would be a nice touch for people who may not fully understand. However, I’m good with leaving it as-is as well.

parkervcp:
Technically I have an egg that you can use to host mysql/mariadb servers.

PR #538, opened 2023-07-29

CHKellerman:
I had to install 'libapache2-mod-php8.1' for apache2 to recognize php.

OS version Ubuntu 20.04


PadowYT2 (commented):
`libapache2-mod-php` should be automatically be picking the latest version you have from repos. Such as currently 8.2, so this means the pr is useless

danny6167:
Thanks for the submission. However, both packages are not required.

PR #539, opened 2023-07-30
Add php8.1-intl as a dependency

LtdJorge:
The initial setup with composer fails otherwise

### DIFF
--- a/panel/1.0/getting_started.md
+++ b/panel/1.0/getting_started.md
@@ -72,7 +72,7 @@ apt update
 apt-add-repository universe
 
 # Install Dependencies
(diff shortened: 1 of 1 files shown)


QuintenQVD0:
it seems like you are the only one having this issue, as nobody else reported it.

I think this can be closed

Boy132:
> it seems like you are the only one having this issue, as nobody else reported it.
> 
> I think this can be closed

Some users on the discord has this problem too.
But I can't remember the exact circumstances/ system setup where this problem occured.

danny6167:
Seems to be a very rare edge case where this needs to be specifically installed, and we haven't seen it in the Discord for a long time.

Thanks for your PR, but we're going to close this one.

PR #545, opened 2023-09-14
update mysql command to mariadb

Jcodeerd:
As seen on this screenshot, taken today, the `mysql` command will be deprecated when using MariaDB image

I do have a few questions though,
- Do I also have to change the references from "MySQL" to "MariaDB" in the guides?
- Should I also change the command for the legacy installation guide? panel/0.7/getting_started.md


Boy132:
As we support both mysql and mariadb I would say it's better to add both commands. Similar to how nginx & apache commands are separated.
Something like:
```
# If using MySQL
mysql -u root -p

# If using MariaDB (version x+ _[whatever version they will rename it]_)
mariadb -u root -p
```


> * Do I also have to change the references from "MySQL" to "MariaDB" in the guides?
> * Should I also change the command for the legacy installation guide? panel/0.7/getting_started.md

No and no

Jcodeerd:
AFAIK there's no specific date given yet, I've put it as v11.0.0+ for now

QuintenQVD0:
Wy not put mariadb first? As that is what the default docs install.

Jcodeerd:
Good point, I don't really think it matters (users should know whether they are using mariadb or mysqld), but I'll change it

panel/1.0/additional_configuration.md
```diff
@@ -127,7 +127,13 @@ reCAPTCHA can easily be disabled using the admin panel. In the Settings, select
 If you cannot access your panel, you can modify the database directly using the following commands.
 
 ```sql
+# If using MariaDB (V11.0.0+)
```
matthewpi, line 130: Please use a lowercase `v` when referring to versions, otherwise my brain will explode :)

Also adding a note stating this is the default when following the docs may be useful so people don't get confused.

Jcodeerd: Resolved in my latest commits

rowansc1 (approved):
I'm happy with these changes personally. 
Facts seem to be correct and in order. Thanks!

Issue #540, opened 2023-08-05
Request: Definite firewall/port requirements in documentation


lowland3r:
Please consider adding a section to the Panel and Wings installation guide that includes the ports required for each component to work (as well as what ports can be access-restricted)

zvikasdongre:
There is a section in which the diagram of all services are shown in docs.
Please have a look at [Pterodactyl Terminology](https://pterodactyl.io/project/terms.html#advanced-setup-diagram)

![image](https://github.com/pterodactyl/documentation/assets/47200577/4963530a-947e-414e-888f-d6df6dff19ec)

jrwimmer:
Indeed, though this doesn't fully clarify what ports on the wings node must be accessible to the panel vs accessible to the world. 

I know that tcp/2022 must be reachable by end users, but what about tcp/8080?

zvikasdongre:
Technically you don't need to make tcp/8080 public, but it would be necessary if you want to monitor the wings on a server.
And making the port tcp/8080 accessible only by the panel's ip will cause the node list to show red heart because the request to node is made from the user's browser, not from the panel, other than that, I think it should be fine.

MinerPL:
Both ports (tcp/8080 and tcp/2022) must be accessible by the end user. Port tcp/8080 is responsible for WebSocket communications as well as downloading files.

zvikasdongre:
Oh yeah, sorry my bad, forgot about websockets and file download connections. 😅

rowansc1:
To be honest, I think that the [terminology ](https://pterodactyl.io/project/terms.html#advanced-setup-diagram)in the docs does a pretty good job.

I don't personally think this needs any further explanation. Please feel free to elaborate if you disagree. 

Perhaps, it may be worth including a link in the installation guide to make it easier to find? I'm open to suggestions!

rowansc1:
Closing due to inactivity.

PR #546, opened 2023-09-22
Defined port requirements

dev-tejas:
## Issue #540 Fixed

Added The Port Requirements In Documentation. As Suggested In Issue #540


QuintenQVD0:
Wtf are those ports? Have you ever used pterodactyl? As there not even close to been right.

PR #550, opened 2023-10-14
Update Community guide for Fedora 38

Lovinoes:
Corrections for the PHP section on Fedora 38.
+
Change "Fedora Server 38" to "Fedora 38" as it does not matter which environment you are working in as they are all the same.
tested on Fedora Cloud, Server, Workstation

### DIFF
--- a/community/installation-guides/panel/centos8.md
+++ b/community/installation-guides/panel/centos8.md
@@ -56,7 +56,7 @@ dnf module enable php:remi-8.1
 dnf update -y
 
 ## Install PHP 8.1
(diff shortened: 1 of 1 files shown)


parkervcp:
While I agree that it isn't specific to Fedora Server i don't think we should encourage people running this on a full desktop machine.

Lovinoes:
I just made some corrections to the documentation.
Nowhere is it mentioned that you can run the panel on a workstation environment.

parkervcp:
I also understand that. By specifying Server it encourages using that instead of the Workstation version. As for the Cloud versions, it's just minimal  images built to run on cloud services. I would expect that to be closely related to the server edition.

Lovinoes:
👌

PR #551, opened 2023-10-14
Add sudo to systemctl enable --now wings

LouissXI:
Every systemctl command in the documentation has sudo before, why not this one ?


Jcodeerd:
It was probably missed in https://github.com/pterodactyl/documentation/pull/548, good catch

danny6167:
Thank you for your contribution. We pulled your commit from this into https://github.com/pterodactyl/documentation/pull/553 and it has merged.

PR #557, opened 2023-10-22
Update php to version 8.2 debian.md

CodeExplorerX:
Update php to version 8.2

### DIFF
--- a/community/installation-guides/panel/debian.md
+++ b/community/installation-guides/panel/debian.md
@@ -31,7 +31,7 @@ echo "deb [signed-by=/usr/share/keyrings/redis-archive-keyring.gpg] https://pack
 apt update -y
 
 # Install PHP and required extensions
-apt install -y php8.1 php8.1-{common,cli,gd,mysql,mbstring,bcmath,xml,fpm,curl,zip}
+apt install -y php8.2 php8.2-{common,cli,gd,mysql,mbstring,bcmath,xml,fpm,curl,zip}
 
 # MariaDB repo setup script
 curl -sS https://downloads.mariadb.com/MariaDB/mariadb_repo_setup | sudo bash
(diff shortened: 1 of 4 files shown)


CodeExplorerX (commented):
Update php to version 8.2

QuintenQVD0:
The panel does not run on PHP 8.2? aleast the queue worker does not

CodeExplorerX:
> The panel does not run on PHP 8.2? aleast the queue worker does not

I tested it locally on Debian 12 and Ubuntu 23.04. Pteodactyl panel runs on php 8.2 as well as a queue worker

QuintenQVD0:
> > The panel does not run on PHP 8.2? aleast the queue worker does not
> 
> I tested it locally on Debian 12 and Ubuntu 23.04. Pteodactyl panel runs on php 8.2 as well as a queue worker

Ubuntu 23.04 is not even an LTS so that to begin with is not supported. 

The queue worker does not work last time we checked. It does not run its crons that the users in the panel setup

QuintenQVD0:
and there are way more places in the docs where php 8.1 is mentioned as we lock everything to a specific version

CodeExplorerX:
> and there are way more places in the docs where php 8.1 is mentioned as we lock everything to a specific version

Okay, I'll try to finalize and attach all the tests

Boy132:
> The panel does not run on PHP 8.2? aleast the queue worker does not

Since the Laravel 10 update it runs fine on php 8.2. The test even use 8.1 & 8.2.

QuintenQVD0:
Please also see #558 and make this pr work with that.

CodeExplorerX:
#558

CodeExplorerX:
#558

CodeExplorerX:
#558

PR #558, opened 2023-10-22
drop support php 8.0

QuintenQVD0:
Drop support for php 8.0 as we updated to Laravel 10 what only supports php 8.1 or 8.2
afbeelding

PR #572, opened 2024-02-15
Update getting_started.md to replace PHP 8.1 with 8.2

jrichman-r7:
PHP 8.1 is no longer available in default repos per my test on Debian 12. However, PHP 8.2 was found.


PadowYT2:
This isn't the only place where PHP 8.1 is mentioned. Also this is a duplicate PR (#557)

Boy132:
There are more places where you would need to change the version. Also, the default guide assumes Ubuntu as base os.

Besides, there is also #557 and #558

jrichman-r7:
my mistake, thanks for your due diligence

PR #575, opened 2024-03-01
Correct caddy PHP_VALUE value as it doesn't support \n

danny6167:
Line breaks in environment variables in caddy fpm configs don't support the \n notation.

This was causing PHP to set `upload_max_filesize` to `100M \n post_max_size = 100M`.

This lead to `upload_max_filesize` being evaluated to an invalid value that causes file uploads (such as egg imports) to fail due to the size being above `upload_max_filesize`.

This PR replaces the \n with an actual newline.

I've confirmed this works in a fresh testing environment.

### DIFF
--- a/.snippets/webservers/Caddyfile
+++ b/.snippets/webservers/Caddyfile
@@ -15,7 +15,8 @@
         root /var/www/pterodactyl/public
         index index.php
 
-        env PHP_VALUE "upload_max_filesize = 100M \n post_max_size = 100M"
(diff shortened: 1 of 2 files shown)


VasilisThePikachu:
Just tested this myself and it works

hexlocation:
+1 had been diagnosing this issue for hours, this fixed it :3

matthewpi (approved):

PR #542, opened 2023-09-03
Use packages for installing NodeJS instead of the installer as not supported

PadowYT2:
As stated on [nodesource/distributions#readme](https://github.com/nodesource/distributions#readme):
## **New Update ⚠️**

We'd like to inform you of important changes to our distribution repository [nodesource/distributions](https://github.com/nodesource/distributions).

**What's New:**

- _**Installation Scripts:** The installation scripts `setup_XX.x` are no longer supported and are not needed anymore, as the installation process is straightforward for any RPM and DEB distro._

---

Closes #543

PR #579, opened 2024-03-29
Update Node.js install script

Jelcoo:
Update the Node.js install script to install with https://github.com/nodesource/distributions


Jelcoo:
Rgarding https://github.com/pterodactyl/documentation/pull/542, it seems like they changed their mind
chrome_OEBZlJgi2P
https://github.com/nodesource/distributions?tab=readme-ov-file#new-update-%EF%B8%8F

PR #569, opened 2023-12-22
Update creating_ssl_certificates.md

Codixer:
Made a seperation of the documentation on CF's side.

PR #580, opened 2024-04-03
fix typo in Resources

Rajdave69:
Ressources -> Resources


Boy132:
Will be fixed in #569

PR #581, opened 2024-04-05
replace redis repository with ppa

alix1383:


QuintenQVD0:
Their docs still tell you to use the keyring?

https://redis.io/docs/install/install-redis/install-redis-on-linux/

Issue #562, opened 2023-11-29
Sidebar top is hidden


PadowYT2:
While using a small device, when you open the sidebar, the top sidebar (with the close button) is hidden by the banner

---

I'll try to fix this

PR #588, opened 2024-06-23
Remove some left over margins from the banner that was just removed

danny6167:
This removes some of the left over margins from the banner that was just removed from the website.
Right now it's an empty space that's just pushing everything down.


danny6167:
This will also resolve https://github.com/pterodactyl/documentation/issues/562

matthewpi (approved):

PR #595, opened 2024-07-29
Update terms.md

samerfarida:
Updated the simple and advanced diagrams  as a code using mermaid

### DIFF
--- a/project/terms.md
+++ b/project/terms.md
@@ -25,11 +25,110 @@ created on nodes, and you can have multiple servers per node.
 
 
 ## Simple Setup Diagram
-![](./../.vuepress/public/simple_setup_diagram.png)
+```mermaid
+flowchart TD
+    Panel[Panel] --> Node1[Node #1]
+    Panel[Panel] --> Node2[Node #2]
+    Panel[Panel] --> Node3[Node #n]
 
+    Node1 ---> Server1[Server #1]
+    Node1 ---> Server2[Server #2]
+    Node1 ---> Server3[Server #3]
+    Node1 ---> Server4[Server #n]
+
+    Server1 --> Nest1[Nest #1]
+    Server1 --> Nest2[Nest #2]
+
+    Nest1 --> Eggs1[Eggs #1]
+
+    Eggs1 --> Yolk11[Yolk #1]
+    Eggs1  --> Yolk12[Yolk #2]  
+```
 
 ## Advanced Setup Diagram
 ::: tip Panel and Wings on the same machine
 It is also possible to install wings on the panel machine so it acts as panel and node machine at once.
 :::
-![](./../.vuepress/public/example_setup.png)
+
+
+
+```mermaid
+flowchart LR
+
+    subgraph P[ Panel Machine ]
+        Redis[(Redis)]
+        Database[(Database<br> 'MySQL/MariaDB')]
+        subgraph Web[Webserver <br> 'Nginx/Apache']
+            Panel[Panel]
+        end
+    end
+
+    subgraph N[Node Machines]
+        Wings[Wings]
+        subgraph Docker[Docker]
+            subgraph C[Container]
+                subgraph Server[Server]
+                end
+            end
+        end
+    end
+
+
+    admin((Administrator))<-->|80 & 443| Web
+    player1((Player #1))<-->|Node IP : Game Port#| Server
+    player2((Player #n))<-->|Node IP : Game Port#| Server
+    Server ~~~ player1 
+    Server ~~~ player2 
+
+    Panel <--> |3306| Database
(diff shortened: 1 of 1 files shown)


samerfarida (commented):
I have reviewed this code. Is there anything else you need from me?

rowansc1:
Sorry for the wait @samerfarida! This will get looked into.

danny6167:
Is there something missing from your PR?
Seems this mermaid stuff isn't something it knows how to render.
https://pr-595.ptero-doc-preview.pages.dev/project/terms.html

samerfarida:
> Is there something missing from your PR? Seems this mermaid stuff isn't something it knows how to render. https://pr-595.ptero-doc-preview.pages.dev/project/terms.html

There is no issue with the current PR itself. However, I wasn't aware that the .md file would be converted to .html, which introduces the rendering problem. After reviewing the Mermaid.js documentation at https://mermaid.js.org/config/usage.html#simple-full-example, it seems that in order to properly render Mermaid diagrams in HTML, the code needs to be wrapped as follows:

```
<!-- Other HTML code -->

<pre class="mermaid">
    <!-- Insert Mermaid code here -->
</pre>
<script type="module">
    import mermaid from 'https://cdn.jsdelivr.net/npm/mermaid@11/dist/mermaid.esm.min.mjs';
</script>

<!-- More HTML code -->

```

For example, I have attached an .html file that demonstrates this. Please download it (rename it by removing the .txt extension) and open it locally in a browser to see a Mermaid diagram rendered within HTML. Below is a sample of the Mermaid code inside the HTML structure:

[mermaidTest.html.txt](https://github.com/user-attachments/files/17237658/mermaidTest.html.txt)

```
<!doctype html>
<html lang="en">
  <body>
    <pre class="mermaid">
      graph LR
      A --- B
      B-->C[fa:fa-ban forbidden]
      B-->D(fa:fa-spinner);
    </pre>
    <script type="module">
      import mermaid from 'https://cdn.jsdelivr.net/npm/mermaid@11/dist/mermaid.esm.min.mjs';
    </script>
  </body>
</html>

```
At this point, we have three possible options moving forward:

1. Save the Mermaid diagrams as .png files and replace the current diagrams. This way, they will render correctly on both GitHub and the website.
2. Modify the build process to wrap the Mermaid code in the required HTML structure during the conversion from .md to .html, as recommended by Mermaid.js.
3. I can update the current PR to include the necessary HTML structure. However, this will prevent the Mermaid code from rendering correctly in the .md file on GitHub, though it will work on the website.

Please let me know which option you and the Pterodactyl team would prefer.

Thanks,
Sammy

rowansc1:
Hi Sammy,

Thank you very much for that detailed analysis. 

Personally, I think option 2 is the most preferable. However, option 3 would also still be fine as the docs are meant to be viewed on the website anyway. 

What do you think @danny6167 ?

danny6167:
I really don't want to see us adding little hacks to the build process for things like this.
In-lining HTML, especially with external dependencies (something we don't do) really isn't a great option either.

pre-rendered images is really the only valid option of the options that's been listed. But the ideal solution would be if there was a native solution for vuepress to be able to render the mermaid syntax.

It really doesn't matter how it looks in github, what matters is how it looks when it's tested after the vuepress build.

samerfarida:
> I really don't want to see us adding little hacks to the build process for things like this. In-lining HTML, especially with external dependencies (something we don't do) really isn't a great option either.
> 
> pre-rendered images is really the only valid option of the options that's been listed. But the ideal solution would be if there was a native solution for vuepress to be able to render the mermaid syntax.
> 
> It really doesn't matter how it looks in github, what matters is how it looks when it's tested after the vuepress build.

Hey @danny6167, I have no problems with pre-rendered images. I did a quick search on https://vuepress.vuejs.org/ and it looks like they have [Markdown plugins](https://marketplace.vuejs.press/plugins/markdown.html) that can be enabled. 

Apparently, they do have a [Mermaid plugin](https://plugin-md-enhance.vuejs.press/guide/chart/mermaid.html) for your VuePress site. Please review. It looks like it’s an easy install and enablement. This way, you guys are set for future usages of diagrams as code! 

Please let me know if you still want me to correct this PR to include a pre-rendered  images? 

Thank you, 

Sammy

danny6167:
That's a good find @samerfarida - I'll take a look at that and update soon if it works out.

danny6167:
Hi!

Just letting you know I haven't forgotten about this.

We can't merge this right now. Looks like we are going to have to update some dependencies, possibly even move to the next version of vuepress to be able to support this properly. That's going to take some time.

Leave the PR open, and we will merge it when the rest of the code is ready for it.

Thank you for your contribution!

PR #596, opened 2024-08-01
panel/1.0/troubleshooting: Add note about cert expiration

suutari:
Expired certificate can create an error that looks like a general communication error, since websocket or XHR requests just fail in very similar way like the service was not responding at all.

This could be a quite common case, especially if the certificate renewal (e.g. for Let's Encrypt) is not properly set up.

### DIFF
--- a/panel/1.0/troubleshooting.md
+++ b/panel/1.0/troubleshooting.md
@@ -70,6 +70,7 @@ and `2022` for SFTP traffic.
 https://domain.com:8080` on the Panel server and ensure that it can successfully connect to Wings.
 * Ensure that you are using the correct HTTP scheme for your Panel and Wings. If the Panel is running over HTTPS
   Wings will also need to be running on HTTPS.
+* If using HTTPS for Wings, make sure that the certificates have not expired.
 
 ### More Advanced Debugging Steps
 * Stop Wings and run `wings --debug` to see if there are any errors being output. If so, try resolving them manually,


danny6167:
Not sure how much this is really going to help.
Two of the existing steps would have already exposed the fact the certificate was expired.

suutari:
You're right that the expired cert would be exposed with e.g. the curl command, but I just thought that maybe explicitly listing this potential issue could still help, since you know what to look for.  And possibly could just have an "aha moment" while reading the troubleshooting list even without checking those previous steps.

suutari:
Also, is issuing the `curl` command from the server the correct way to check the connectivity? At least in my case the issue was when the browser wasn't able to connect to the 8080 port via websocket, because of expired cert. But I don't really know anything about Pterodactyl to touch the existing bullet points in that list.

suutari:
If you think this note is worthless feel free to close the ticket.

danny6167:
I didn't say it's worthless, I just want to have real and open discussions about how much of an effect it's really going to have.

We'll be taking a look into these smaller PRs after the major docs update PR is merged.

suutari:
Sorry, didn't mean to be dramatic. And thanks for letting me know about the plan!

The truth is that I found the expired cert with the help of the previous bullet points. My thoughts at the time were that this kind of common issue could be mentioned explicitly still to speed up the debugging process, but that's only speculation of course.

Another thing that probably could have helped even more would have been a copy-paste from the error message that is shown in the UI: `We're having some trouble connecting to your server, please wait...` or the one that is shown in the browser's console: `Firefox can’t establish a connection to the server at wss://...:8080/api/servers/...`. I think users start debugging by searching the web with those error messages.

Let me know if you want me improve the troubleshooting steps even more with those. I can update this PR or create another.

rowansc1:
Honestly, the vast amount of people who have issues due to cert expiry could be stopped by this, but they don't usually realise until they get shown first hand, but I'm not against merging this in. Thoughts @danny6167?

And @suutari thanks for your PR! Please feel free to make more in the future. If you're unsure if its needed, feel free to make an issue first explaining your enhancement before you spend the time editing things.

danny6167:
They should already be discovering this issue with the existing steps, but I'm not opposed to this being merged.
It's simple, short, might make people think a little more, and doesn't cause an additional confusion. :)

rowansc1 (approved):
+1 thanks for the PR

PR #607, opened 2024-09-29
Update to centos8.md community guide

Arkanuki:
I added the php-process package containing the posix php module required by Composer to the dependencies list.

I was blocked because Composer was throwing
`Root composer.json requires PHP extension ext-posix * but it is missing from your system. Install or enable PHP's posix extension.`

The posix extension needed is part of the php-process package. Adding that package fixed the problem.


rowansc1:
It's probably fine to implement, I'm going to replicate the issue first though.

PR #614, opened 2024-10-09

echumley:
Using the instructions above this code block, you create a Cloudflare API Token and not an API Key. The API token is the only variable you need to export before running the acme.sh command. You don't need to include an API Key or email and will work without it.


PadowYT2:
Using `CF_Token` is only for the modern acme.sh (including `CF_Account_ID` & `CF_Zone_ID`). The legacy is legacy and *shouldn't* be edited

PR #616, opened 2024-10-17
Update getting_started.md

azfoxd:
Add Caddy reference to the permissions part, otherwise the panel will return a 500 error if using Caddy as a webserver when trying to access it because permissions weren't applied.

Not sure if it is necessary on RHEL, Alma/Rocky as well as I mainly use Debian.

### DIFF
--- a/panel/1.0/getting_started.md
+++ b/panel/1.0/getting_started.md
@@ -184,7 +184,7 @@ The last step in the installation process is to set the correct permissions on t
 use them correctly.
 
 ``` bash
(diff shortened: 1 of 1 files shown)


rowansc1:
> > CC: @danny6167 - IIRC you use Caddy?
> 
> We use it on our frontend. Those servers run Debian and the caddy process runs as user `caddy` by default, but with the way our docs work the php-fpm process would still be running under `www-data` on Debian based platform.
> 
> We can probably merge this. At worse it's not any less correct.

Cool cheers. Will merge now.

rowansc1 (approved):
👍

PR #622, opened 2024-11-21
Update updating.md

barbabarros:
Added troubleshooting for 502 Bad Gateway after updating php.


rowansc1 (changes requested):
I'm not a fan of this approach, especially hardcoding PHP v8.3 in there (for future upgradability).

Another thing, I don't think this is the best way to go about it, maybe just modifying the existing note about PHP would make it more compact. 

Give it another go and experiment, as this isn't a bad idea. 

Thanks for your PR!

PR #623, opened 2024-11-23
Future-proof php version in schedules troubleshooting guide

Loki-101:
Previous "up to" isn't specific enough now that there's only two supported versions, and I may as well use the page that's built for this exact purpose if I'm updating it.

### DIFF
--- a/panel/1.0/troubleshooting.md
+++ b/panel/1.0/troubleshooting.md
@@ -140,8 +140,8 @@ If this returns different DNS Servers than 1.1.1.1 and 1.0.0.1 you'll need to ed
 - Check logs from your queue manager ``journalctl -xeu pteroq``
 - Restart pteroq ``systemctl restart pteroq``
 - Clear schedule cache ``php /var/www/pterodactyl/artisan schedule:clear-cache``
-- Check your php version ``php -v`` - [this page](<https://pterodactyl.io/panel/1.0/updating.html#panel-version-requirements>) will tell you what versions of php are supported by what versions of the panel
-- Check your crontab syntax using <https://crontab.guru/> - make sure it's what you intended
+- Check your php version ``php -v`` - [this page](https://pterodactyl.io/panel/1.0/updating.html#panel-version-requirements) will tell you what versions of php are supported by what versions of the panel
+- Check your crontab syntax using https://crontab.guru/ - make sure it's what you intended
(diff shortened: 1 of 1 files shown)


rowansc1 (changes requested):
Line 143 is incorrect, the `<>` are not needed in the context you are using and should be removed. 

The `<>` are used independently when you want to pop something out which is just a link, not an embedded link, such as in the contab guru link below. 

Example: 
```
[this page](https://pterodactyl.io/panel/1.0/updating.html#panel-version-requirements)
```

Thanks for your PR!

panel/1.0/troubleshooting.md
```diff
@@ -140,7 +140,7 @@ If this returns different DNS Servers than 1.1.1.1 and 1.0.0.1 you'll need to ed
 - Check logs from your queue manager ``journalctl -xeu pteroq``
 - Restart pteroq ``systemctl restart pteroq``
 - Clear schedule cache ``php /var/www/pterodactyl/artisan schedule:clear-cache``
-- Check your php version - up to 8.3 is supported ``php -v``
+- Check your php version ``php -v`` - [this page](<https://pterodactyl.io/panel/1.0/updating.html#panel-version-requirements>) will tell you what versions of php are supported by what versions of the panel
```
rowansc1, line 143: Line 143 is incorrect, the `<>` are not needed in the context you are using and should be removed. 

The `<>` are used independently when you want to pop something out which is just a link, not an embedded link, such as in the contab guru link below. 

Example: 
```
[this page](https://pterodactyl.io/panel/1.0/updating.html#panel-version-requirements)
```

Thanks for your PR!

rowansc1 (changes requested):
Line 143 is incorrect, the `<>` are not needed in the context you are using and should be removed. 

The `<>` are used independently when you want to pop something out which is just a link, not an embedded link, such as in the contab guru link below. 

Example: 
```
[this page](https://pterodactyl.io/panel/1.0/updating.html#panel-version-requirements)
```

Thanks for your PR!

Loki-101:
Whoops, too used to markdown for Discord. I removed the <> for crontab.guru while I was at it; that was leftover from when I was sending this guide in Discord before the original PR was accepted.

Loki-101:
Whoops, I was cleaning up old repositories and forgot I had this PR open. I'll make a new one.

PR #625, opened 2024-12-11
Future proof php version in docs

Loki-101:
Continuation of https://github.com/pterodactyl/documentation/pull/623

PR #642, opened 2025-01-26
Update updating.md

Red-Banana-Official:
The documentation is slightly misleading in the supported php versions. It correctly states to use 8.2 or 8.3 at the top of the page, but scrolling down it says ''...running atleast 8.2'' indicating that 8.4 might also be supported (which it isn't)

Changed it to say ''...running atleast 8.2 or 8.3'' to avoid some confusion

Thanks to https://github.com/DataDropp for noticing it

### DIFF
--- a/panel/1.0/updating.md
+++ b/panel/1.0/updating.md
@@ -34,8 +34,8 @@ There are no 1.8.x, 1.9.x, or 1.10.x releases of Wings.
 - PHP `8.2`, or `8.3` (recommended)
 - Composer `2.X`
 
-**Before continuing**, please ensure that your system and web server configuration has been upgraded to at least PHP 8.2 by running `php -v` and Composer 2 by running `composer --version`. You
-should see an output similar to the result below. If you do not see at least PHP 8.2 and Composer 2, you will need to upgrade by following
+**Before continuing**, please ensure that your system and web server configuration has been upgraded to at least PHP 8.2 or 8.3 by running `php -v` and Composer 2 by running `composer --version`. You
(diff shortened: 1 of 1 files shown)


panel/1.0/updating.md
```diff
@@ -34,8 +34,8 @@ There are no 1.8.x, 1.9.x, or 1.10.x releases of Wings.
 - PHP `8.2`, or `8.3` (recommended)
 - Composer `2.X`
 
-**Before continuing**, please ensure that your system and web server configuration has been upgraded to at least PHP 8.2 by running `php -v` and Composer 2 by running `composer --version`. You
-should see an output similar to the result below. If you do not see at least PHP 8.2 and Composer 2, you will need to upgrade by following
+**Before continuing**, please ensure that your system and web server configuration has been upgraded to at least PHP 8.2 or 8.3 by running `php -v` and Composer 2 by running `composer --version`. You
```
rowansc1, line 37: "at least PHP 8.2 or 8.3" still implies that PHP 8.4 is supported. Could you reword this?

panel/1.0/updating.md
```diff
@@ -34,8 +34,8 @@ There are no 1.8.x, 1.9.x, or 1.10.x releases of Wings.
...
+should see an output similar to the result below. If you do not see at least PHP 8.2 or 8,3 and Composer 2, you will need to upgrade by following
```
rowansc1, line 38: Same here

rowansc1 (changes requested):
Thanks for the PR! Easy rewords to do :)

panel/1.0/updating.md
rowansc1: Actually, it might be worth following the same method as in #625. That would future proof this from being incorrect in the future.

PR #644, opened 2025-02-02
Add Issue Templates & Config

itzMiney:
Added a config for issues disallowing empty issues and linking to the Pterodactyl Discord
Also added a templates using forms for clarification requests and enhancements

grafik

Edit: Maybe I should've checked how forms work more thoroughly before pushing to master ggs

### DIFF
--- a/panel/1.0/getting_started.md
+++ b/panel/1.0/getting_started.md
@@ -137,15 +137,13 @@ php artisan key:generate --force
 ```
 
 ::: danger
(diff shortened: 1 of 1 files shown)


rowansc1:
Reopened

rowansc1:
I think this is a nice edit. Thoughts @danny6167?

danny6167:
Is clarification really any different to an improvement?
I'd also say we should have a an option for reporting something actually wrong with the docs.

PR #649, opened 2025-02-10
Update installing.md

Nartek01:
I was having trouble finding how to access the panel view. So I've added a little warning, telling that to be able to access the panel simply type the address 127.0.0.1

### DIFF
--- a/wings/1.0/installing.md
+++ b/wings/1.0/installing.md
@@ -112,6 +112,9 @@ set when creating the node.
 :::
 
 ## Configure
+::: Accessing Panel
+In your url type 127.0.0.1
+:::
 
 Once you have installed Wings and the required components, the next step is to create a node on your installed Panel. Go to your Panel administrative view, select Nodes from the sidebar, and on the right side click Create New button.
 


Nartek01:
@itzMiney What I meant is that there's nothing mentioning in the guide on how to actually access the panel. I think this is worth mentioning.

image

danny6167:
Thank you for the attempt here, but this just isn't even valid in 99% of setups.
We make the fairly reasonable assumption that users know the address of what ever machine they have installed Pterodactyl on.
Teaching people the basics of IP addressing is beyond the scope of our documentation.

Nartek01:
@DaneEveritt Understandable, thank you.

itzMiney:
> @itzMiney What I meant is that there's nothing mentioning in the guide on how to actually access the panel. I think this is worth mentioning.

@Nartek01 that is because the environment setup command in the [Getting Started](https://pterodactyl.io/panel/1.0/getting_started.html#environment-configuration) Guide is pretty clear at least in my opinion
grafik
If you set the application URL to the Loopback IP or a Private IP then that is of course gonna be where you will be able to reach it

PR #652, opened 2025-02-13
feat: add tip about swap enabled by default on linux kernel v6.1 and newest

WeeskyBDW:
this is just a tip as danny6167 pointed, not all distribution use a 6.1 or newest kernel

### DIFF
--- a/wings/1.0/installing.md
+++ b/wings/1.0/installing.md
@@ -75,7 +75,9 @@ sudo systemctl enable --now docker
 ```
 
 #### Enabling Swap
-
+::: tip Recent Linux kernel
+Since the version 6.1 of the linux kernel, swap is enabled by default. If you are running a kernel version 6.1 or newer, you can skip this step. To check your kernel version, run `uname -r`.
+:::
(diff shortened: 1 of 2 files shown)


wings/1.0/installing.md
```diff
@@ -75,7 +75,9 @@ sudo systemctl enable --now docker
 ```
 
 #### Enabling Swap
-
+::: tip Recent Linux kernel
+Since the version 6.1 of the linux kernel, swap is enabled by default. If you are running a kernel version 6.1 or newer, you can skip this step. To check your kernel version, run `uname -r`.
+:::
```
PadowYT2, line 80:
```suggestion
::: tip Recent Linux kernel
Since the version 6.1 of the linux kernel, swap is enabled by default. If you are running a kernel version 6.1 or newer, you can skip this step. To check your kernel version, run `uname -r`.
:::

```

danny6167 (approved):

PadowYT2 (commented):
lgtm, maybe some capitalization changes (like Linux)

PR #654, opened 2025-03-04
Permission issue mitigation tip

SoNotMC:
Just a helpful tip, fair enough if you do not merge this tho!


rowansc1:
Thank you for your help! As much as this is true, people installing this software should have a basic understanding of Linux, and thus know to use it as root or using `sudo`. 
Most commands in the docs which require `sudo` have this included anyway. 
Most people who install this have a root account by default anyway, this isn't a very common issue so I'm not going to merge this in. Thanks for the PR though!

PR #657, opened 2025-03-07
Update mysql_setup.md

itzMiney:
Changed it to mention the pterodactyl0 interface IP from wings instead of docker's docker0 IP


Jcodeerd (commented):
There’s a small typo here: pterodacrtyl0 instead of pterodactyl0

rowansc1 (changes requested):
Thanks for the PR! Some changes to be made, but thank you for trying to improve pterodactyl!

rowansc1 (approved):
Yup happy with that! Thank you :)

PR #659, opened 2025-04-12
Update outdated line

Loki-101:
/home/container support for mounts has been available for a while now, and the section above this explicitly mentions it. This line must have been missed.


rowansc1 (approved):
You are correct! Thank you very much for that nice edit. I'll merge this in! :)

PR #660, opened 2025-04-28
Correct php8.3-pdo to php8.3-common

danners430:
php8.3-pdo automatically resolves to php8.3-common when installed - propose bringing the documentation in line to help when troubleshooting


rowansc1 (approved):
Thanks for the PR @danners430!

Good thinking! You are correct here. pdo is replaced by common now days. Cheers!

PR #661, opened 2025-05-01
Added two "::: warning"s for two problems that I encountered while installing + One other minor edit

kaz-49:
I encountered some problems while installing Pterodactyl Panel on Fedora Server 41, so I figured I'd add the solutions I found here. 
I'm quite sure that these apply to more RHEL/Fedora-based distros, but I'm afraid I haven't tested installing the Panel on other distros so I don't know. 

I also changed the preexisting "::: tip" to a "::: warning", because I think that the information in that "::: tip" block sounds much more important than your average (potentially optional) tip.

### DIFF
--- a/community/installation-guides/panel/centos8.md
+++ b/community/installation-guides/panel/centos8.md
@@ -79,6 +79,6 @@ sudo systemctl enable --now php-fpm
 Excellent, we now have all of the required dependencies installed and configured. From here, follow the [official Panel installation documentation](/panel/1.0/getting_started.md#download-files).
 
 
-::: warn
+::: warning
 You will need to change the fastcgi_pass path in the Nginx configuration to `/var/run/php-fpm/pterodactyl.sock`
 :::
--- a/community/installation-guides/panel/centos8.md
+++ b/community/installation-guides/panel/centos8.md
@@ -79,6 +79,6 @@ sudo systemctl enable --now php-fpm
 Excellent, we now have all of the required dependencies installed and configured. From here, follow the [official Panel installation documentation](/panel/1.0/getting_started.md#download-files).
 
 
-::: tip
+::: warn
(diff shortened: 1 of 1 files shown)


Jcodeerd:
I wonder whether its better to add these commands to [the troubleshooting page](https://github.com/pterodactyl/documentation/blob/ff449b91381f7674e18c58bde93649d844545b29/panel/1.0/troubleshooting.md?plain=1#L105)?

kaz-49:
I think with this one (below), it should be fine to just add it to the troubleshooting page as you said. 
> ::: warning
> If you complete the [official Panel installation documentation](/panel/1.0/getting_started.md#download-files) but get a HTTP 500 error code and "Permission Denied" errors in `/var/log/nginx/pterodactyl.app-error.log`, you may need to run the following commands: 
> 
> ``` bash
> sudo chcon -R -t httpd_sys_rw_content_t storage
> sudo chcon -R -t httpd_sys_rw_content_t bootstrap/cache
> ```
> :::

With this one (below), though, the command that I have and the commands that are on the troubleshooting page both appear to be solutions for similar (or perhaps even the same) problems. I don't know which of these solutions are better security-wise, so it would be great to get some input from someone who actually knows SELinux. If we can't get ahold of that person though, I think it'd be fine to remove mine because I think (based on nothing but gut feeling) that the pre-existing one looks more secure. 
> ::: warning
> If you get "Permission denied \[tcp://127.0.0.1:6379\]" in your laravel log (`/var/www/pterodactyl/storage/logs/laravel-TODAYS-DATE.log`), you may need to run the following commands: 
> 
> ``` bash
> sudo setsebool httpd_can_network_connect=1
> ```
> :::

I want to keep the edit from `::: tip` to `::: warning` as-is because of the aforementioned reason though. 
> ::: warning
> You will need to change the fastcgi_pass path in the Nginx configuration to `/var/run/php-fpm/pterodactyl.sock`
> ::: 


Should I edit the PR to make these changes?

PR #669, opened 2025-06-09
Update panel.md

ayu2606:
change node version to 20 as 16 is decrepted


WeeskyBDW:
It would be better to use Node 22 as Node 20 will pass in maintenance mode in october. Node 22 will be supported until 2027

ayu2606 (commented):
Changed to 22

rowansc1:
Thanks for the PR! Merged this in :)

PR #676, opened 2025-07-02
Update creating_ssl_certificates.md to fix the certbot auto renewal c…

thecrazytech1:
…ommand for crontab

I've added --nginx to the certbot renew command as I and multiple other people have had issues with that command not working due to nginx running and using port 80 when it tries to renew the domains


danny6167:
certbot renew will automatically use the same method that was used to create the certificate in the first place.
If you had to add --nginx to get it to renew then you must have generated it standalone.

Really that entire section should be removed. As far as I'm aware, on all supported platforms the installation of certbot will automatically install a crontab/systemd timer to auto renew.

PR #683, opened 2025-07-14
Update troubleshooting.md

itzMiney:
I rewrote the first step in advanced troubleshooting steps to use `sudo wings diagnostics` to stop people from overreading the **Stop Wings** part and creating duplicate instances of Wings. Most useful errors for debugging get logged on a normal run anyways so the diagnostics command makes more sense in my opinion.

### DIFF
--- a/panel/1.0/troubleshooting.md
+++ b/panel/1.0/troubleshooting.md
@@ -73,7 +73,7 @@ https://domain.com:8080` on the Panel server and ensure that it can successfully
 * If using HTTPS for Wings, make sure that the certificates have not expired.
 
 ### More Advanced Debugging Steps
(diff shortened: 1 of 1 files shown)


panel/1.0/troubleshooting.md
```diff
@@ -73,7 +73,7 @@ https://domain.com:8080` on the Panel server and ensure that it can successfully
 * If using HTTPS for Wings, make sure that the certificates have not expired.
 
 ### More Advanced Debugging Steps
-* Stop Wings and run `wings --debug` to see if there are any errors being output. If so, try resolving them manually,
```
rowansc1, line 76: Maybe amend to keep the top line, but with a caveat of if wings are NOT running as a system service? I like the second line though so can keep this!

rowansc1 (changes requested):

PR #684, opened 2025-07-15
Supported os update

itzMiney:
I added some examples of Operating Systems that are commonly tried and NOT supported to make it more clear that the software will not work right on them (or might work but no support will be provided for it)


rowansc1:
Thanks for the PR!
Arch _could_ work, so I'm going to add a quick comment in that.

itzMiney:
> @itzMiney thoughts on this change before I merge?

sure seems like a good ammendment, we could also ammend that Debian 13 works too with downgraded dependencies maybe?

rowansc1:
Yeah sure! Debian 13 only just came out afaik, so we could see some more people wanting to do that.

PR #702, opened 2025-09-26

Kasova-code:
Added a note in both “Nginx With SSL” and “Nginx Without SSL” sections explaining that on Ubuntu systems it may be necessary to run `sudo usermod -aG www-data nginx` to ensure proper permissions.

### DIFF
--- a/panel/1.0/webserver_configuration.md
+++ b/panel/1.0/webserver_configuration.md
@@ -16,6 +16,14 @@ First, remove the default NGINX configuration.
 rm /etc/nginx/sites-enabled/default
 ```
 
+::: tip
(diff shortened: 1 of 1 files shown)


Kasova-code:
In my case, Nginx was running under the nginx user, which wasn’t part of the www-data group. Because of that, it couldn’t access the PHP-FPM socket, and I ran into a 503 error after following the setup instructions.

Adding the nginx user to the www-data group resolved the issue. As a beginner, this was unexpected and not obvious at first, so I thought it might be helpful to mention this edge case for others who might encounter the same problem.

Let me know if you'd prefer this info to be added elsewhere or need more details.

PR #717, opened 2025-12-23
Remove legacy docs for panel and nodejs daemon

danny6167:
This removes docs for the legacy versions of the panel and daemon.

Anyone who hasn't updated by this point really needs to be looking at doing a clean install, and the existence of these regularly confuse users.


danny6167:
Included in https://github.com/pterodactyl/documentation/pull/728

